PR #162, opened 2015-01-06
Autocomplete

lucidstack:
autocompletion pp

This PR introduces a first version of the autocompletion feature for the search, present on all pages.

The functionalities are pretty basic at the moment, only supporting autocompletion in the `LIKE 'SAN%'` form. This means that `Guil` won't match `San Guillermo`, but only protected areas and countries starting with `guil` (case insensitive).


Gemfile.lock
```diff
@@ -135,6 +135,7 @@ GEM
     ffi (1.9.6)
     gdal (0.0.5)
     hike (1.2.3)
+    hitimes (1.2.2)
```
adammulligan: Where are we using hitimes?

adammulligan:
nice one, it works really well!

We do, however, need some styling for the smaller search forms :grinning:

screenshot 2015-01-08 12 06 12
screenshot 2015-01-08 12 06 21

Gemfile.lock
lucidstack: We aren't, directly. It's a sidekiq dependency, as far as I'm aware.

lucidstack:
Thank you! 

Yes, the styling is definitely not the best. I opened the PR knowing we needed to talk about CSS :grinning:

PR #174, opened 2017-06-01
New site page styling

stacytalbot:


app/assets/javascripts/modules/tabs.js.coffee
```diff
@@ -1 +1,25 @@
+define('tabs', [], ->
+  class Tabs
+    constructor: (@$container) ->
+      @addEventListners()
```
lucidstack: You've got a tiny typo here! 😀

app/assets/javascripts/modules/tabs.js.coffee
```diff
@@ -1 +1,25 @@
...
+    addEventListners: ->
+      @$container.find('.js-tab-title').on('click', (e) =>
+        $clicked = e.target
+
+        @updateActiveTabTitle($clicked);
+        @updateActiveTabContent($clicked);
+      )
+
+    updateActiveTabTitle: (tabTitle) ->
+      @$container.find($('.js-tab-title')).removeClass('tab__title--active')
```
lucidstack: I think we can avoid instantiating the jquery object `$('.js-tab-title')`, and just use `find('.js-tab-title')`. Should yield the same result!

app/assets/javascripts/modules/tabs.js.coffee
```diff
@@ -1 +1,25 @@
...
+      @$container.find(tabTitle).addClass('tab__title--active')
+
+    updateActiveTabContent: (id) ->
+      tabContentSelector = '[data-id="' + $(id).data('id') + '-content"]'
```
lucidstack: with coffeescript you get string interpolation (all praise)! 

So this becomes `"[data-id='#{$(id).data('id')}-content']"`

app/assets/javascripts/modules/tabs.js.coffee
```diff
@@ -1 +1,25 @@
...
+
+      @$container.find($('.js-tab-content')).addClass('u-hide')
```
lucidstack: Here the same as the comment above, about the jquery object 😀

PR #185, opened 2017-06-21
New site page amends

stacytalbot:
Updates as per Michels latest design
- Added map that shows the site on its own
- Refactored the site networks map and key


Levia:
I've just spotted a minor issue when switching between networks (when there are many).
If I uncheck the sites on a network, than click on another network and coming back to the previous one, the box are still unchecked but the sites overlays are showing (see screenshot below).
screen shot 2017-06-21 at 16 02 11

Other than that everything looks great! :)

PR #192, opened 2017-07-07
Add show/hide button to the site page so that only 6 protected areas …

stacytalbot:
…are shown initially.

### DIFF
--- a/app/assets/javascripts/modules/protected_areas/base.js.coffee
+++ b/app/assets/javascripts/modules/protected_areas/base.js.coffee
@@ -4,7 +4,9 @@ $(document).ready( ->
   )
 
   require(['show_hide'], (ShowHide) ->
-    new ShowHide($('.js-show-hide'), 6)
+    quota = 6
+
+    new ShowHide($('.js-show-hide'), quota)
   )
 
   require(['map', 'fullscreen'], (Map, Fullscreen) ->
--- a/app/assets/javascripts/modules/protected_areas/base.js.coffee
+++ b/app/assets/javascripts/modules/protected_areas/base.js.coffee
@@ -4,7 +4,7 @@ $(document).ready( ->
   )
 
   require(['show_hide'], (ShowHide) ->
-    new ShowHide($('.js-show-hide'))
+    new ShowHide($('.js-show-hide'), 6)
   )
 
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/modules/show_hide.js.coffee
```diff
@@ -0,0 +1,52 @@
...
+      $wrappers = @$button.parents('.js-show-hide-wrapper')
+
+      $.each($wrappers, (index, wrapper) =>
+        $wrapper = $(wrapper)
+
+        $cards = $wrapper.find('.js-show-hide-target')
+
+        @hideCards($cards)
+
+        if($cards.length < 7)
```
Levia: Can we define a constant variable instead of using a plain integer?
I think it can also come in handy in a piece of code below when you do the slice

app/assets/javascripts/modules/show_hide.js.coffee
```diff
@@ -0,0 +1,52 @@
...
+
+        else
+          @hideCards($cards)
+
+          $clicked.data('type', 'show')
+          $clicked.text('Show all')
+      )
+
+    hideCards: (cards) ->
+      $cardsToHide = cards.slice(6)
```
Levia: See comment above

app/assets/javascripts/modules/show_hide.js.coffee
stacytalbot: Yep, good idea. Updated!

app/assets/javascripts/modules/protected_areas/base.js.coffee
```diff
@@ -4,7 +4,7 @@ $(document).ready( ->
   )
 
   require(['show_hide'], (ShowHide) ->
-    new ShowHide($('.js-show-hide'))
+    new ShowHide($('.js-show-hide'), 6)
```
Levia: It's a good idea to pass this value as a parameter as well. ;)
What I also mean is not to pass/use values as plain integer, but it might be a good idea to define a constant variable that contains that number.
It encourages code recycling and readability, so everybody knows exactly what that number represents :)

stacytalbot: 👍🏻

PR #196, opened 2017-07-20
Disallow robots to index our pages to avoid them to slow down pp

Levia:
Adds a value to the header of the response to help preventing robots to index our pages

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -57,6 +57,8 @@ default: &default
 
   show_pame_statistics: false
 
+  disallow_all_web_crawlers: <%= ENV["DISALLOW_ALL_WEB_CRAWLERS"] %>
+
 development:
   <<: *default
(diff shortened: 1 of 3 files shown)


config/secrets.yml
```diff
@@ -57,6 +57,8 @@ default: &default
 
   show_pame_statistics: false
 
+  disallow_all_web_crawlers: <%= ENV["DISALLOW_ALL_WEB_CRAWLERS"] %>
```
spencerldixon: Dont forget to add the env to `.env.example` with an example of what needs to go in there

spencerldixon: Actually I think if this is just going to be a true/false thing, we could just put it in the secrets as it's nothing sensitive, no need for an environment variable. What do you think?

Levia: Yeah, let's keep it simple. Done

PR #200, opened 2017-08-23
Generate stats for marine page

Levia:
Calculate stats for the marine page and format those that are static

### DIFF
--- a/app/controllers/marine_controller.rb
+++ b/app/controllers/marine_controller.rb
@@ -1,5 +1,7 @@
 class MarineController < ApplicationController
 
+  before_action :coverage
+  before_action :most_protected_areas, only: [:index]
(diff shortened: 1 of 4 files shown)


app/controllers/marine_controller.rb
```diff
@@ -1,6 +1,17 @@
 class MarineController < ApplicationController
 
+  #Calculated stats
+  before_action :coverage
```
spencerldixon: Does a before_action make sense if we only have an index view here?

Levia: It will make the code more structured I believe.
The other way of doing it is just calling the methods inside the action, but supposing we are going to add something in the index action or even create other actions, this will keep things cleaner

PR #211, opened 2017-10-12
Add overseas_territories csv file and importer

Levia:
This adds and overseas territories importer that will updated those records automatically for every new WDPA release.
The migration wasn't enough to cover new WDPA release imports


lib/modules/wdpa/overseas_territories_importer.rb
```diff
@@ -0,0 +1,30 @@
...
+  def import
+    ActiveRecord::Base.transaction do
+      csv = CSV.read(OVERSEAS_TERRITORIES_CSV)
+      csv.shift # remove headers
+
+      csv.each do |parent_iso, child_isos|
+        parent_country = Country.find_by_iso_3(parent_iso)
+        child_isos = child_isos.split(';')
+
+        unless parent_country.nil?
```
lucidstack: you can do a `next if parent_country.nil?` here, and save an indentation level 😀

lib/modules/wdpa/overseas_territories_importer.rb
```diff
@@ -0,0 +1,30 @@
...
+          child_isos.each do |child_iso|
+            child_country = Country.find_by_iso_3(child_iso)
+
+            unless child_country.nil?
```
lucidstack: aaand same here! 😝

PR #215, opened 2017-10-18
Template dropdown

stacytalbot:


app/views/cms/_child_dropdown.html.erb
```diff
@@ -0,0 +1,20 @@
...
+@options = []
+
+@cms_page.children.published.each do |child|
+
+  object = {}
+
+  object[:title] = child.label
+  object[:text] = cms_block_content_render(:text, child)
+  object[:url] = cms_block_content(:url, child)
+  object[:image] = cms_block_content(:image, child).file().url(:dropdownImage) if(cms_block_content(:image, child).try(:file) && cms_block_content(:image, child).try(:file).url(:dropdownImage))
```
Levia: You should be able to get rid of the `try(:file)` on the right and side of the `&&` condition, as it won't check that one if the first condition fails :)

PR #223, opened 2017-11-24
Twitter sharing

stacytalbot:
Add ability to share a highlighted selection of text on Twitter on via email.
Add a sticky tab telling the user they can do the above.


Levia:
This looks great! The only issue I found was that if I try to send the select text via email, the text is not copied over, while is perfectly fine with Twitter.

PR #224, opened 2017-11-28
Add marine stats importer

drewpotter:
I have made the necessary changes to import the marine statistics into Redis and retrieve them as a hash. However, I do feel like I could go further and I could consolidate the two different statistics methods in the controller into just one structure which has all of the necessary data.

### DIFF
--- a/app/controllers/marine_controller.rb
+++ b/app/controllers/marine_controller.rb
@@ -1,8 +1,7 @@
 class MarineController < ApplicationController
 
   #Static stats
-  before_action :total_coverage, only: [:index]
-  before_action :distributions, only: [:index, :download_designations]
+  before_action :marine_statistics, only: [:index, :download_designations]
   before_action :growth, only: [:index]
   before_action :ecoregions, only: [:index]
   before_action :pledges, only: [:index]
@@ -52,7 +51,7 @@ class MarineController < ApplicationController
     @coverageOfTop20ProtectedAreas = [
       {
         title: "Total global coverage of all MPAs",
-        km: @distributions[:nationalWatersKm] + @distributions[:highSeasKm]
+        km: @marine_statistics["protected area coverage of national waters km2"].to_i + @marine_statistics["protected area coverage of the high seas km2"].to_i
       },
       {
         title: "Total global coverage of largest 20 MPAs",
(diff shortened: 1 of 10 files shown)


app/controllers/marine_controller.rb
```diff
@@ -145,21 +144,8 @@ def designations
...
-    @distributions = {
-      nationalWaters: 39,
-      nationalWatersPa: 16.02,
-      nationalWatersKm: 22624944,
-      highSeas: 61,
-      highSeasPa: 0.25,
-      highSeasKm: 558116
-    }
+  def marine_statistics
+    @marine_statistics = Wdpa::MarineStatsImporter.stats
```
Levia: Since we are importing data into Redis, there is no need to call the importer in the controller.
You can still add this importer to the list of importers called in the `wdpa/importer.rb` file.
So the only thing left to do in the controller is retrieving those data from Redis

Levia: I would probably retrieve the stats using a different module or just with a plain Redis call rather than using the importer. It can be confusing, also because "getting" stats doesn't sound like a function of an "importer"

drewpotter: Great point @Levia! I will make an amendment.

Levia:
It looks good overall. I will amend the template of the csv and ask the Protected Area Programme if they like it.
Ideally we want more columns rather than rows; also, the attributes name should be with underscores instead of spaces.
So I won't merge this until we have got a better csv and we have implemented the necessary amendments

drewpotter:
Many thanks @Levia I have made the amendment with regards to Redis and getting those statistics directly. Once the CSV is altered I will need to adjust the names to remove the spaces and have underscores instead within the code.

app/views/marine/index.html.erb
```diff
@@ -2,7 +2,7 @@
...
 <%= content_for :social_image, image_url('social-marine.png') %>
-<%= content_for :social_image_alt, "A map of the world with highlighted areas that represent protected areas. The following statistics are shown: there are 15,271 marine protected areas, #{@oceanProtectedAreasPercent} of the ocean is covered by protected areas and the total area protected is 23,036,078km² These statistics are for 2017." %>
+<%= content_for :social_image_alt, "A map of the world with highlighted areas that represent protected areas. The following statistics are shown: there are #{@marine_statistics["total_marine_protected_areas"]} marine protected areas, #{@marine_statistics["total_ocean_pa_coverage_percentage"]} of the ocean is covered by protected areas and the total area protected is 23,036,078km² These statistics are for 2017." %>
```
Levia: Can we please use single quotes here rather then double quotes while getting the value from the hash? So it's less confusing as the entire content is surrounded by double quotes already.

Levia:
Looks good!
Can you please write some simple tests for the importer?
You should be able to get some hints from other tests in `test/unit/wdpa/`

app/views/marine/index.html.erb
drewpotter: Sure I will update that now @Levia

drewpotter:
Thanks! I will write some tests.

PR #225, opened 2017-12-12
Initial support for the region page with some statistics already in p…

drewpotter:
…lace. I have set up some of the statistics, although perhaps this could be improved somehow? Also I fixed a small part of `app/views/country/_info_box.html.erb` to remove two spaces in the text. There are more statistics which need adding.


app/controllers/region_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      region_hash[:percentage_pa_ts_cover] += country.statistic.percentage_pa_ts_cover || 0
+      region_hash[:land_area] += country.statistic.land_area || 0
+      region_hash[:percentage_pa_cover] += country.statistic.percentage_pa_cover || 0
+      region_hash[:pa_eez_area] += country.statistic.pa_eez_area || 0
+      region_hash[:pa_ts_area] += country.statistic.pa_ts_area || 0
+      region_hash[:percentage_pa_marine_cover] += country.statistic.percentage_pa_marine_cover || 0
+      region_hash[:marine_area] += country.statistic.marine_area || 0
+      region_hash[:polygons_count] += country.statistic.polygons_count || 0
+      region_hash[:points_count] += country.statistic.points_count || 0
+    end
```
Levia: Instead of creating an hash with all the possible values and then going through each attribute to initialise them like this in the controller, you could create a model that calculates just the needed  stats.
For example, you can have a look at the `CountryStatistic` model.
In the model then, you can add functions to calculate only the needed stats; we don't necessarily need to replicate the entire structure of the country (for now at least)

drewpotter: @Levia originally I had planned to create a model called `RegionStatistic` which would store all of the necessary statistics inside a database table. I had originally created a migration for this, but went against it after some discussion, because I thought maybe I could use Redis for this. However, should I create a migration and a table for these statistics called `region_statistics`?

Levia: Sorry, I mistaken `CoutryPresenter` with `CountryStatistic`.
With countries we are given a file with the stats and we import those every month.
With regions, we just need to do some calculations over those country stats; that said, I think it should be enough to have a `RegionPresenter` that uses the `CountryStatistic` model for each country the region contains.
So no need for migrations as `RegionPresenter` in not an ActiveRecord model.
Does this make sense to you?

drewpotter: Thanks @Levia I totally understand what you mean!

config/routes.rb
```diff
@@ -35,6 +35,11 @@
   get '/country/:iso/compare(/:iso_to_compare)', to: 'country#compare', as: 'compare_countries'
   get '/country/:iso/protected_areas', to: 'country#protected_areas', as: 'country_protected_areas'
 
+  get '/region/:iso', to: 'region#show', as: 'region'
+  get '/region/:iso/pdf', to: 'region#pdf', as: 'region_pdf'
+  get '/region/:iso/compare(:/iso_to_compare)', to: 'region#compare', as: 'compare_regions'
+  get '/region/:iso/protected_areas', to: 'region#protected_areas', as: 'region_protected_areas'
```
Levia: I think we can just have the show page route for now. If needed we can add those later, but the main goal of this piece of work is just the show page for the moment

drewpotter: Thanks I will fix this too.

app/presenters/region_presenter.rb
```diff
@@ -94,22 +94,28 @@ def sources_per_jurisdiction
...
-          }
-        }
-        protected_areas_per_iucn_category_array << hash
+    valid_iucn_categories = IucnCategory.all.pluck(:name, :id).uniq
+    region_data = Hash.new { |hash, key| hash[key] = Hash.new }
+    processed_data = []
+
+    region.countries.each do |country|
+      country.protected_areas_per_iucn_category.each do |protected_area|
+        region_data["#{protected_area['iucn_category_name']}"][:count] = Array.new if region_data["#{protected_area['iucn_category_name']}"][:count].nil?
```
Levia: It looks good!
What I would probably do to further improve the readability of this function, is to create a variable in the loop to store the value of `protected_area['iucn_category_name']`.
In this way you don't need to access the protected_area hash every time and it's easier to read with just a variable.

Also, I would replace `Array.new` with just `[]`

drewpotter: Thanks @Levia I will update the code in the way you have suggested!

app/controllers/region_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  before_filter :load_vars
+
+  def show
+  end
+
+  private
+
+  def load_vars
+    @region = Region.where(iso: params[:iso].upcase).first
+
```
spencerldixon: Oh god please get rid of the blank lines

drewpotter: Thanks @spencerldixon I have removed those blank lines.

PR #227, opened 2017-12-18
Added countries count statistic to region show page

spencerldixon:


drewpotter:
Looks really good! Both in the actual region page and also in the code itself.

Levia:
Perfect! Thanks!

PR #228, opened 2017-12-19
Added governance types statistics and created a shared partial

drewpotter:
Please note, I used the same variable name `region_pa_category = region_data[protected_area["governance_name"]]` as a previous method. I am happy to change this variable name if it is not appropriate.

Also the `search_url` appears to me to be valid, but the search results are currently empty.


app/controllers/country_controller.rb
```diff
@@ -49,6 +49,7 @@ def load_vars
 
     @country or raise_404
 
+    @location = @country
```
Levia: The idea of the shared variable name between countries and regions is to go in the partial as a `locals`. 
When you render a partial you can pass variables like so: `render partial: 'view_name', locals: {location: @country}`
That means you are rendering the partial and passing it a variable named location.
Then you can use that variable just like so: `location.method_name`

This means you don't need this line here or in the regions_controller

drewpotter: Ah yes, of course! I will fix this now.

app/views/region/facts/_governance_types.html.erb
```diff
@@ -0,0 +1,26 @@
...
+            <td class="table__value"><%= row["governance_type"] %></td>
+            <td class="table__value"><%= row["count"] %></td>
+            <td class="table__value"><%= row["percentage"].to_f.round(2) %></td>
+          </tr>
+        <% end %>
+        </tr>
+      </tbody>
+    </table>
+  </div>
+</section>
```
Levia: If you used a shared partial I guess you can remove this file.

drewpotter: Hmm, thanks @Levia I deleted that file but it looks like I didn't remove correctly using git. It doesn't appear in my git repo, or when I do `git status`

Levia: Uhm, it probably should appear in git status in a different section; so you then need to do git add for the deleted file as well. Don't worry you can't manage to delete it from the repo, I can do it for you.

app/models/region.rb
```diff
@@ -27,6 +27,30 @@ def protected_areas_with_iucn_categories
...
+  def protected_areas_per_governance
+    region_data = Hash.new { |hash, key| hash[key] = Hash.new }
+    processed_data = []
+
+    countries.each do |country|
+      country.protected_areas_per_governance.each do |protected_area|
+        region_pa_category = region_data[protected_area["governance_name"]]
+        region_pa_category["governance_type"] = protected_area["governance_type"] if region_pa_category["governance_type"].nil?
+        region_pa_category["count"] = [] if region_pa_category["count"].nil?
+        region_pa_category["count"] << protected_area["count"].to_i
```
Levia: You can also probably improve this and the percentage by doing: `region_pa_category["country"] += protected_area["count"].to_i` meaning that you are already summing up the values here; so you don't need to do the `reduce` afterwards

drewpotter: Thanks @Levia, should I also do this for the `region_pa_category["percentage"]` ?

Levia: Yep, it should work the same way. Just remember to divide by the count where you are already doing that

Levia:
Well done!

PR #229, opened 2017-12-19
Designation stats

spencerldixon:


app/views/region/facts/_designations.html.erb
```diff
@@ -0,0 +1,48 @@
...
+        <div class="space-between">
+          <div class="fact__link">
+          </div>
+          <div class="fact__link">
+          </div>
+        </div>
+      <% end %>
+    </div>
+  </div>
+</div>
```
Levia: Do you think we could use a shared partial between regions and countries? Or would you prefer doing this along with all the others in a later PR?

spencerldixon: There's a slight difference; `@country` vs `@region` but we can abstract this out. I would say let's merge it for now and then have a card to do it later

PR #231, opened 2017-12-20
Fixed percentages for iucn management categories

drewpotter:
I had to fix the `sources_per_jurisdiction_hash` like I did for my other pull request because I needed to get the page loading to fix this other issue. There might be an issue with merging. Anyway the percentages have been fixed now!


Levia:
Cool! We need to speak with Brian about the percentages and if we need to display more decimal numbers for very low percentages

PR #233, opened 2017-12-20
Temporarily disable downloads on region page

drewpotter:
I have commented out the sections for the downloads. There might be a better more elegant way of commenting this out, but I went with the suggestion I found on Stack Overflow.


Levia:
To comment out the HTML, the correct way is surrounding the lines with `<!-- -->`

PR #234, opened 2017-12-21
Merge partials for region and country statistics

drewpotter:
All of the partials have been merged together into a shared folder apart from those which cannot be merged. Such as `_related_countries.html.erb` and `_number_of_countries.html.erb`.

One outstanding question is the `_pame_statistics.html.erb` which has not been included in the `region` page. This could always be added if necessary.


app/models/region.rb
```diff
@@ -53,6 +53,52 @@ def protected_areas_per_governance
...
+        region_pa_category["count"] ||= 0
+        region_pa_category["count"] += protected_area["count"].to_i
+        total_region_count << protected_area["count"].to_i
+        region_pa_category["percentage"] ||= 0
+        region_pa_category["percentage"] += protected_area["percentage"].to_f
+      end
+    end
+
+    processed_data = region_data.map{ |key,value| {
+          "iucn_category_name" => key,
```
spencerldixon: Spacing!

app/models/region.rb
```diff
@@ -53,6 +53,52 @@ def protected_areas_per_governance
...
+          "count" => value["count"],
+          "percentage" => 100 * value["count"] / total_region_count.reduce(0, :+)
```
spencerldixon: It's preferred to use symbols for the hashes over strings where possible https://stackoverflow.com/questions/8189416/why-use-symbols-as-hash-keys-in-ruby

app/models/region.rb
```diff
@@ -53,6 +53,52 @@ def protected_areas_per_governance
     }
   end
 
+  def protected_areas_per_iucn_category
+    region_data = Hash.new { |hash, key| hash[key] = Hash.new }
```
spencerldixon: This hash is unnecessary as far as I can see, it just ends up creating an empty hash. Might as well use the literal `region_data = {}`

app/models/region.rb
```diff
@@ -53,6 +53,52 @@ def protected_areas_per_governance
...
+    total_region_count = []
+
+    countries.each do |country|
+      country.protected_areas_per_iucn_category.each do |protected_area|
+        region_pa_category = region_data[protected_area["iucn_category_name"]]
+        region_pa_category["count"] ||= 0
+        region_pa_category["count"] += protected_area["count"].to_i
+        total_region_count << protected_area["count"].to_i
+        region_pa_category["percentage"] ||= 0
+        region_pa_category["percentage"] += protected_area["percentage"].to_f
```
spencerldixon: Seems like there would be an easier way to sum the counts rather than looping over everything and adding them up one by one. Perhaps inject? I'm not sure what ruby version we're on either, but 2.4 has this http://ruby-doc.org/core-2.4.0_rc1/Enumerable.html#method-i-sum

app/models/region.rb
spencerldixon: Or do we even need to define empty arrays and hashes? Can we just make them as we go?

drewpotter: @spencerldixon I did try making the hash as we go but I got some errors, so I found this solution to the problem on Stack Overflow. I could try and make the change you suggest but I get the following result:

```
undefined method `[]' for nil:NilClass
```

app/models/region.rb
drewpotter: Originally I tried to use `inject` in the first version of this method. But the code got really complex and had to be re-mapped again afterwards. So this was the rewrite of that. I will look into this suggestion @spencerldixon thanks!

app/models/region.rb
drewpotter: Sorry about the spacing! Fixed it.

app/models/region.rb
drewpotter: Thanks @spencerldixon the only thing is that if we choose to use symbols we need to rewrite the other parts of the code which are already using strings. I went this path so we could unify the partials. If you want we can rewrite the other code which was already there, but we need to do that if we choose to use symbols in the same shared partial.

app/models/region.rb
Levia: instead of that line, have you tried with just `region_data = {}` or did it fail as well?

drewpotter: Thanks @Levia I will try that now!

drewpotter: When I try that I get:

`undefined method `[]' for nil:NilClass`

app/models/region.rb
```diff
@@ -53,6 +53,52 @@ def protected_areas_per_governance
...
   end
 
+  def protected_areas_per_iucn_category
+    region_data = Hash.new { |hash, key| hash[key] = Hash.new }
+    processed_data = []
+    total_region_count = []
+
+    countries.each do |country|
+      country.protected_areas_per_iucn_category.each do |protected_area|
+        region_pa_category = region_data[protected_area["iucn_category_name"]]
```
Levia: I think the problem lies here.
The `region_data` hash at this point is always empty, so it fails to retrieve the data inside `region_data[protected_area["iucn_category_name"]]` 
I'm not sure what your intent was here, but maybe you just need `protected_area["iucn_category_name"]` ?

drewpotter: Thanks @Levia it was originally different, but it got changed to make it more concise. The purpose of that line:

```
region_pa_category = region_data[protected_area["iucn_category_name"]]
```

Was to make that protected_area's `iucn_category_name` the key inside the `region_data` hash. Which is then referenced to by the `region_pa_category` in subsequent lines.

I will try your suggestion!

PR #237, opened 2018-01-08
Put iucn categories in the correct order

drewpotter:
As requested, I have ordered the categories in the desired order.

### DIFF
--- a/app/models/region.rb
+++ b/app/models/region.rb
@@ -66,16 +66,16 @@ class Region < ActiveRecord::Base
       end
     end
 
-    processed_data = correct_order.map do |key,value|
+    correct_order.each do |key,value|
       next if region_data[key].nil?
-      {
+      processed_data << {
         "iucn_category_name" => key,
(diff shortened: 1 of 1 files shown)


app/models/region.rb
```diff
@@ -65,14 +66,16 @@ def protected_areas_per_iucn_category
...
+      {
+        "iucn_category_name" => key,
+        "count" => region_data[key]["count"],
+        "percentage" => 100 * region_data[key]["count"] / total_region_count.reduce(0, :+)
       }
-    }
+    end
 
-    processed_data
+    processed_data.compact
```
Levia: Do we need `.compact` here? Ideally we are skipping all `nil` values in the previous loop

drewpotter: True @Levia! I don't know why but I got nil values coming through, the loop must need changing to remove this `compact`.

drewpotter: It appears the `next if` is outputting a nil value when that occurs.

PR #238, opened 2018-01-08
Remove reference from countries page

spencerldixon:


Levia:
I wonder if we need to remove this from the regions page as well. I'll check with Brian and comment that card

PR #240, opened 2018-02-27
Initial work towards updating the documentation with regards to an error

drewpotter:
This adds a certain error which may appear and how to fix it.


drewpotter:
@stacytalbot thanks for looking at this PR 😃 I think it is good to be merged. Unless @Levia would like me to move this recommendation from the readme to somewhere else in the docs?

Levia:
Not sure about the force-pushes either, but it should be about rebasing this branch to master.
Happy for this to be merged

PR #241, opened 2018-04-17
Fixed search results to correctly display the pa coverage total percentage

drewpotter:
This is the initial work to fix the search results to correctly display the PA coverage total percentage.

### DIFF
--- a/app/views/search/_country.html.erb
+++ b/app/views/search/_country.html.erb
@@ -7,6 +7,7 @@
   </h3>
 
   <p>
-    <%= StatisticPresenter.new(model).percentage_pa_cover %>% covered by Protected Areas
+    <% stats = StatisticPresenter.new(model) %>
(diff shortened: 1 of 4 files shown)


app/views/search/_country.html.erb
```diff
@@ -7,6 +7,7 @@
   </h3>
 
   <p>
-    <%= StatisticPresenter.new(model).percentage_pa_cover %>% covered by Protected Areas
+    <% stats = StatisticPresenter.new(model) %>
+    <%= (stats.percentage_pa_land_cover + stats.percentage_pa_marine_cover).round(2) %> % covered by Protected Areas
```
Levia: Could you please add this logic into the Presenter module?
It would be good to keep the views agnostic in terms of any logic, even for a simple one like the sum

drewpotter: Sure @Levia great point!

PR #245, opened 2018-06-25
Changed pame evaluations method attribute to methodology to avoid res…

spencerldixon:
…erved words clash


Levia:
Is this missing a migration perhaps? The `PameEvaluation` model still has a `method` column

spencerldixon:
Work just needs the pame model attributes changing from method to methodology, as method is a reserved word and doesn't play nicely with the tests.

This work is on hold for the time being so will leave the PR open to be picked up in the future. Pame evaluations should NOT show in the API atm.

PR #248, opened 2018-08-17
Change PAME subnav URL and add external icon

stacytalbot:
## Description

Create the option to add in an external URL to a subnav. Roger is aware this isn't the best solution but happy with this implementation for now.

You will need to update the GD-PAME page in your CMS to test.

screen shot 2018-08-20 at 11 33 07


drewpotter:
This works fine for me, merging into master :)

PR #250, opened 2018-09-20
Use wdpa id on search results

Levia:
## Description

Some protected areas have the same slug/name, so search could be inefficent sometimes especially when then going to the show page.
That will work with the wdpa id instead of the slug now.


drewpotter:
Looks good to me!

PR #255, opened 2018-11-28
Feature/content for malaysia

Levia:
### Description

* Add management plan documents for Malaysia in:
  - Country page
  - Specified PAs

### Known issues

It's a bespoke feature for Malaysia to be delivered as soon as possible.
If becomes more popular, this should implemented properly and for more a more general purpose.

### DIFF
--- a/app/helpers/countries_helper.rb
+++ b/app/helpers/countries_helper.rb
@@ -10,12 +10,12 @@ module CountriesHelper
   def malaysia_documents
     [
       {
-        url: 'url1',
-        name: 'name1'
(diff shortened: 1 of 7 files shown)


app/helpers/countries_helper.rb
```diff
@@ -0,0 +1,22 @@
+module CountriesHelper
+  def is_malaysia?
+    @country && @country.iso_3 == "MYS"
+  end
+
+  def is_japan?
```
drewpotter: @Levia I spotted this `is_japan` method and was just curious about that, since everything else seems to be focused on Malaysia. Otherwise I am happy to merge!

PR #256, opened 2019-01-21
Banner

stacytalbot:
HTML changes to add the Live Report link into the primary navigation.


WKocur:
Sure - shall I pop this on a card somewhere or assume it will be covered when the redesign of the header is done?

PR #257, opened 2019-01-21
Added monthly import helper

spencerldixon:
Added a monthly import helper to make importing the stats easier

### DIFF
--- a/lib/modules/import_tools.rb
+++ b/lib/modules/import_tools.rb
@@ -22,4 +22,14 @@ module ImportTools
   def self.dump_path
     Rails.root.join("lib", "data", "seeds", "pre_seeded_database.sql")
   end
+
+  def self.monthly_import
+    ActiveRecord::Base.transaction do
+      CountryStatistic.delete_all
+      PameStatistic.delete_all
+      Stats::CountryStatisticsImporter.import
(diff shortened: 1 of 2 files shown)


lib/modules/import_tools.rb
```diff
@@ -22,4 +22,14 @@ def self.last_import
   def self.dump_path
     Rails.root.join("lib", "data", "seeds", "pre_seeded_database.sql")
   end
+
+  def self.monthly_import
```
Levia: Is this used anywhere? 
Not sure if it's implicitly called somewhere.

Also, what about the previous code in the importer itself?

spencerldixon: We run this every month to import the statistics, I've just added it as a helper here so that I can just call `ImportTools.monthly_import` instead of writing out 5 lines in the console each time. Would be happy to move it elsewhere if needs be or work it into the process to automate it where possible

Levia: Makes sense. I was thinking that perhaps the `Wdpa::PameImporter` could be added into the list of importers of the `execute_importers` method inside the `Wdpa::Importer` class.

Sorry, too many `importers`!

PR #258, opened 2019-01-30
Banner

WKocur:
This works for one slide, but would need a little bit adding for multiple slides showing at one time (but hopefully not too much!)


app/assets/javascripts/home.js
```diff
@@ -1,2 +1,7 @@
+//= require 'vue'
 //= require base
+//= require 'vue'
```
stacytalbot: vue has been required twice here

PR #262, opened 2019-03-05
Add story map model

lucacug:


lib/modules/wdpa/story_map_link_list_importer.rb
```diff
@@ -8,18 +8,8 @@ def import
...
-        else
-          pa = ProtectedArea.find_by_slug(row[0])
-        end
-
-        unless pa.blank?
-          pa.story_map_link = row[1]
-          pa.save
-          puts "Added this link #{row[1]} to #{row[0]}"
-        end
+        StoryMapLink.where(wdpa_id: row[0], link: row[1])
```
Levia: This will create a new record even if there's no ProtectedArea with that WDPA ID.
Can we please check if a PA exists before adding story links?
If a PA does not exist, it would be good to log it on the screen

lucacug:
@stacy we are still waiting for Marine to say if it is ok to add to the csv file a field (something like `type`) to distinguish between multiple story link for the same wdpa

app/models/protected_area.rb
```diff
@@ -9,6 +9,7 @@ class ProtectedArea < ActiveRecord::Base
   has_many :networks_protected_areas, dependent: :destroy
   has_many :networks, through: :networks_protected_areas
   has_many :pame_evaluations
+  has_many :story_map_links, primary_key: :wdpa_id, foreign_key: :wdpa_id
```
Levia: You don't need the `primary_key` definition here.
The `foreing_key` is the one that will handle the relationship

app/models/story_map_link.rb
```diff
@@ -0,0 +1,3 @@
+class StoryMapLink < ActiveRecord::Base
+  belongs_to :protected_area, primary_key: :wdpa_id, foreign_key: :wdpa_id
```
Levia: Same here

app/models/protected_area.rb
Levia: But to be honest I think you can remove the foreign_key as well and use the default id to handle the relationship.
Even though the links are related to the wdpa_id, here we are working with database objects which have also a unique id for each wdpa_id. So specifying a different foreign_key it's probably unnecessary.

You can also check how other relationships are made within PP

PR #263, opened 2019-03-12
Remove everything related to kml format download

lucacug:


lucacug:
@stacy All fixed now!!

PR #265, opened 2019-03-13
Feature/blog

stacytalbot:
## Description

- Add layout file for the blog page
- Add months to the blog page filter
- Remove Updates & News from the primary nav and replace with link to Blog

## CMS Setup

Screenshot 2019-03-13 at 15 31 35

Screenshot 2019-03-13 at 15 31 53


app/helpers/cms_helper.rb
```diff
@@ -35,4 +35,41 @@ def cms_page_path page
...
+
+  def breadcrumb_link page
+    "<a href='/c#{page.full_path}' title='Visit #{page.label}' class='breadcrumbs__link'>#{page.label}</a>"
+  end
+
+  def get_breadcrumbs current_page
+    page = current_page
+    breadcrumbs = []
+
+    breadcrumbs.push(breadcrumb_link(page))
```
Levia: When there is only one argument passed in to the array, I would use `<<` rather than `push`.
So `breadcrumbs << breadcrumb_link(page)`

app/helpers/cms_helper.rb
```diff
@@ -35,4 +35,41 @@ def cms_page_path page
...
+    
+    while hasParent?(page) && page.parent.label != 'Index'
+      page = page.parent
+
+      breadcrumbs.push(breadcrumb_link(page))
```
Levia: Same here

PR #268, opened 2019-03-26
Add message to country pages that have restricted protected areas

stacytalbot:


app/helpers/countries_helper.rb
```diff
@@ -7,6 +7,12 @@ def is_japan?
     @country && @country.iso_3 == "JPN"
   end
 
+  def has_restricted_sites?
+    restricted_iso3 = ["RUS", "EST", "CHN", "GBR"]
+    
+    return restricted_iso3.include? @country.iso_3
```
Levia: `return` is not strictly necessary here, the function will automatically return the result of the latest line of code.
 
Moreover, you could include the `@country` bit here and call this only function in the view

app/views/shared/stats/facts/_coverage_stats.html.erb
```diff
@@ -101,10 +101,16 @@
       countries due to difference in methodologies and datasets used to assess
       protected area coverage and differences in the base maps used to measure
       terrestrial and marine area of a country or territory.
+      
       <% if @country && @country.iso_3 == 'JPN' %>
         Download <a href="https://www.cbd.int/doc/world/jp/jp-nr-05-en.pdf" class="link-with-icon">Japan’s fifth national report (pdf).</a>
       <% end %>
     </p>
+    <% if @country && has_restricted_sites? %>
```
Levia: As mentioned above, if you include the `@country &&` bit in the helper function it would be better

app/views/shared/stats/facts/_coverage_stats.html.erb
```diff
@@ -101,10 +101,16 @@
       countries due to difference in methodologies and datasets used to assess
       protected area coverage and differences in the base maps used to measure
       terrestrial and marine area of a country or territory.
+      
       <% if @country && @country.iso_3 == 'JPN' %>
```
Levia: Also noticed here it could be `is_japan?` as it's defined in the helper

app/views/shared/stats/facts/_coverage_stats.html.erb
stacytalbot: done

app/helpers/countries_helper.rb
stacytalbot: done

app/views/shared/stats/facts/_coverage_stats.html.erb
stacytalbot: done

PR #270, opened 2019-04-11
set permitted params for search helper

WKocur:
This is just a start as I think 'sanitisation' will be needed elsewhere, but my main questions are:

1) does it seem ok to use permit like this
2) am I using the right list of permissible parameters (from Search.rb)
3) Why is params an argument for clear_filters_link (wouldn't they be accessibly in the method under `params` anyway?)
4) The two commented out links in `.../search/index.html.erb` are not visible on pp? What should they do/should they be visible?


WKocur:
Ideally you could set the permitted params on an action/controller basis...

PR #274, opened 2019-04-18
Feature/temporal wdpa

WKocur:


app/controllers/historic_wdpa_controller.rb
```diff
@@ -0,0 +1,13 @@
+class HistoricWdpaController < ApplicationController
+  def index
+    wdpa_releases = HistoricWdpaRelease.all.order(:year, :month).as_json
+
+    wdpa_releases.each do |release|
+      release[:name] = "#{release['year']} - #{Date::MONTHNAMES[release['month']]}"
```
lucacug: It looks very good, I would only change this to `release['name']` for consistency.

lucacug (commented):
@WKocur  I can't see the PR on the PP db repo related to the migration you have added

WKocur:
> @WKocur I can't see the PR on the PP db repo related to the migration you have added

Oops, because there wasn't one! I've made one now. I'm assuming it should be with develop...

app/controllers/historic_wdpa_controller.rb
WKocur: Cheers - will change that now :)

lucacug (commented):
Looks good now from the BE side. Happy to merge if @stacytalbot has already checked the FE bits...

PR #285, opened 2019-05-13
Marine updates

stacytalbot:
Deployed to staging.
Please can you review and if all ok I will get sign off from client before we merge into develop.

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/28


WKocur:
Also there's a failing test, although it doesn't seem related... Is this a known failing test?

WKocur:
Re Ed's comment - I think the wrong bit has been made with 2dps

WKocur:
@stacytalbot All seems to work - I'm happy to approve this if Ed is happy with it. There are a couple of places where I might have more/fewer decimal places, but best to let Ed have the call on that.

PR #286, opened 2019-05-15
Feature/Add restricted protected areas flag

drewpotter:
This is the initial work to add a boolean flag `restricted` to the `PameEvaluation` model. As well as creating a needed join table between `pame_evaluations` and `countries`, as well as modifying the import module for this change.

(cherry picked from commit 623dd4951e5d0c8f44011cd79a5ef92acef8e869)


lib/modules/wdpa/pame_importer.rb
```diff
@@ -29,16 +32,17 @@ def self.import
           ps.language   = row[12]
         end
 
-      if protected_area.nil?
-        missing_pas << wdpa_id
-      else
-        PameEvaluation.where({
+      if protected_area.nil? && (restricted == false) # If PameEvaluation doesn’t have a PA and isn’t restricted it should be deleted.
```
Levia: Could you please add the relevant comments just before the line they refer to rather then within the same line? 
Also, when it's just about one atomic boolean condition, I feel like parenthesis are not really necessary to encapsulate that.

Both are for readibility reasons.
Same in the `elsif` condition below. Exception made for the first group of paranthesis which are encapsulating boolean conditions correctly.

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "db"]
   path = db
   url = https://github.com/unepwcmc/protectedplanet-db.git
-  branch = master
+  branch = feature/add-restricted-protected-areas-flag
```
Levia: Could you please revert this back to the main branch?
This would only need to be changed locally

lib/modules/wdpa/pame_importer.rb
drewpotter: Thanks @Levia I will fix this!

.gitmodules
drewpotter: Thanks @Levia I will change this too!

PR #289, opened 2019-05-17
Frontend updates

stacytalbot:


drewpotter:
@stacytalbot I took a quick look and it seems like webpacker is crashing out for me.
```
webpacker_1      | /usr/bin/env: 'ruby': No such file or directory
protectedplanet_webpacker_1 exited with code 127
```
I will try and get this fixed asap first thing next week.

PR #296, opened 2019-07-15
Feature/add restricted protected areas flag

drewpotter:
This is the (hopefully) final work towards fixing the restricted protected areas flag.


test/unit/wdpa/pame_importer_test.rb
```diff
@@ -33,6 +33,6 @@ class TestPameImporter < ActiveSupport::TestCase
     assert_equal 9, pame_evaluations.count
 
     hidden_pame_evaluation = PameEvaluation.find(64)
```
Levia: Where is the `64` coming from?
I would suggest creating a more dynamic solution for tests. 
For example, instead of fetching an evaluation and check if `protected_area_id` is `nil`, you could try to fetch pame evaluations where the protected_area_id is nil and then assert the expected count of the result of that query.

PR #308, opened 2019-09-06
T11 table

stacytalbot:
- Add table component
- Add table head component
- Loose styling
- Add format for table data

Screenshot 2019-09-09 at 13 24 14

### DIFF
--- a/app/assets/stylesheets/_base.scss
+++ b/app/assets/stylesheets/_base.scss
@@ -1,6 +1,6 @@
-.new-styles { //temporary until the refresh work happens
-  color: $body-font-color; //temporary until the refresh work happens
-  font-family: $body-font; //temporary until the refresh work happens
+.new-styles { //temp-refresh
+  color: $body-font-color; //temp-refresh
+  font-family: $body-font; //temp-refresh
 
   * {
     box-sizing: border-box;
@@ -16,9 +16,9 @@
   h1 { @include h1; }
   h2 { @include h2; }
   
-  p,a { color: inherit; } //can remove when old stylesheets are removed in refresh
+  p,a { color: inherit; } //temp-refresh
   
-  a:link, //override old stylesheets - can removed when refresh is done
+  a:link, //temp-refresh
   a { 
     text-decoration: underline; 
 
--- a/app/assets/stylesheets/_base.scss
+++ b/app/assets/stylesheets/_base.scss
@@ -13,6 +13,16 @@
   h1 { @include h1; }
   h2 { @include h2; }
 
+  section { 
+    color: inherit; //can remove when old stylesheets are removed in refresh
+    font-family: inherit; //can remove when old stylesheets are removed in refresh
+  } 
+
+  div { 
+    color: inherit; //can remove when old stylesheets are removed in refresh
+    font-family: inherit; //can remove when old stylesheets are removed in refresh
+  }
+
   p,a { color: inherit; } //can remove when old stylesheets are removed in refresh
   
   a:link, //override old stylesheets - can removed when refresh is done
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/_base.scss
```diff
@@ -13,6 +13,16 @@
...
+  } 
+
+  div { 
+    color: inherit; //can remove when old stylesheets are removed in refresh
+    font-family: inherit; //can remove when old stylesheets are removed in refresh
+  }
+
   p,a { color: inherit; } //can remove when old stylesheets are removed in refresh
   
   a:link, //override old stylesheets - can removed when refresh is done
```
WKocur: Shall we use a consistent comment for this so that it's easy to find and delete when that happens?

stacytalbot: yea fair point - can i sort out at the end?

stacytalbot: @WKocur ok done!

app/controllers/target_dashboard_controller.rb
```diff
@@ -6,52 +6,78 @@ def index
...
-                title: 'Global', value: 55, target: 75, colour: 'global' 
-              }
-            ]
-          }
-        ] 
-      }.to_json
-    # ]
+
+    ###
+    # need to get text into yml files
```
WKocur: Is this a todo? I usually mark todo as TODO: (because I have an extension that picks that up, but also easy to search by if always the same and codebase will pick this up)

app/javascript/components/table/TableHead.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <div class="table__head">
+    <div v-for="(heading, index) in headings"
```
WKocur: Would span be more appropriate here? I usually think of div as full width or not positioned within the natural flow of the page and then span as inline. Not sure if I'm right there.

app/controllers/target_dashboard_controller.rb
```diff
@@ -6,52 +6,78 @@ def index
...
+              title: 'Areas of importance for biodiversity',
+              charts: [
+                { 
+                  title: 'Global', value: 55, target: 75, colour: 'global' 
+                }
+              ]
+            }
+          ] 
+        },
+        {title:'Spain',url:'http://localhost/country/spain',stats:[{title:'Coverage',charts:[{title:'Terrestrial',value:55,target:75,colour:'terrestrial'},{title:'Marine',value:55,target:75,colour:'marine'}]},{title:'Effectivelymanaged',charts:[{title:'Terrestrial',value:55,target:75,colour:'terrestrial'},{title:'Marine',value:55,target:75,colour:'marine'}]},{title:'Wellconnected',charts:[{title:'Global',value:55,target:75,colour:'global'}]},{title:'Areasofimportanceforbiodiversity',charts:[{title:'Global',value:55,target:75,colour:'global'}]}]}
```
WKocur: urls don't work, but as it's just dummy data I would't worry about fixing.

WKocur:
@stacytalbot Just the two little comments above, then I'm happy :)

app/controllers/target_dashboard_controller.rb
stacytalbot: removed

app/javascript/components/table/TableHead.vue
stacytalbot: They will need to contain the sort button so don't want it in a span

app/controllers/target_dashboard_controller.rb
stacytalbot: Yea i made them up

app/javascript/components/table/TableHead.vue
stacytalbot: They will most likely be display flex not block or inline

PR #310, opened 2019-09-09
Make sure data is ordered as requested

Levia:
## Description

Order global stats data as showed in the design.
Coverage -> Effectively managed -> Representative -> Well connected -> Importance


Levia:
@stacytalbot There are no extra steps.
Do you mean the order is not as desired? Or is it a different issue?

PR #313, opened 2019-09-11
Implement search

stacytalbot:
## Frontend

- Add in searchable select list 
- Add sticky bar
- infinite scroll (load more items when you get to the bottom of the page)

## Not included

- Back end functionality for the search


app/assets/stylesheets/application.scss
```diff
@@ -154,16 +154,20 @@
...
 
 @import './map/custom-leaflet';
 
-@import './helpers/*';
-
-@import './components/*';
\ No newline at end of file
+@import './styles/base';
+@import './styles/helpers/**/*';
+@import './styles/components/icons';
```
WKocur: Is there a reason these are written out individually?

app/assets/stylesheets/styles/components/_carousel.scss
```diff
@@ -139,8 +145,8 @@ $transition-options: 0.6s ease-in-out;
 
     .carousel-slide {
       @include shadow--grey-light;
-      @include responsive-styles(width, 100%, calc(40% - 5px), calc(20% - 10px));
-      @include responsive-styles(margin, 0, rem-calc(0 5), rem-calc(0 10));
+      @include responsive-styles(width, 100%, calc(40% - #{$carousel-basic-slide-margin-tablet}px), calc(20% - #{$carousel-basic-slide-margin-desktop}px));
```
WKocur: Why not set as rem and then use rem inside calc as well?

app/assets/stylesheets/styles/components/form/_select.scss
```diff
@@ -0,0 +1,124 @@
...
+$select-width-desktop: rem-calc(500);
+$select-height: rem-calc(48);
+
+$select-padding-side: rem-calc(12);
+$select-padding: rem-calc(10) $select-padding-side;
+
+$search-icon-padding-right: rem-calc(18);
+$search-icon-padding-left: $select-padding-side;
+
+$select-input-padding-right: calc(#{$search-icon-padding-right} + #{$search-icon-padding-left} + #{$icon-search-width});
```
WKocur: are these all rem? Don't need calc if that is the case

app/javascript/components/table/VTable.vue
```diff
@@ -1,50 +1,50 @@
...
-    />
-
     <table-row 
       v-for="(row, index) in items"
       :key="getVForKey('row', index)"
       :row="row"
     />
-  </div>  
+
+    <span v-if="triggerElement" :class="[ triggerElement, { 'icon-visible': isLoading }, 'icon--loading-spinner margin-center' ]"></span>
```
WKocur: I would extract this class - getting a bit much for the template

app/javascript/mixins/mixin-select-shared.js
```diff
@@ -0,0 +1,143 @@
...
+        switch (e.keyCode) {
+        case KEYCODES.down:
+          this.incremementKeyboardFocus()
+          break
+        case KEYCODES.up:
+          this.decrementKeyboardFocus()
+          break
+        case KEYCODES.enter:
+          console.log('enter')
+          if(this.filteredOptions.length) {
```
WKocur: Can this part of the switch be extracted into a separate function? Also if you reckon the select should work like this normally can you add a ticket to the comp lib to update it? Cheers!

app/assets/stylesheets/application.scss
stacytalbot: So that they are imported in the correct order

app/javascript/components/table/VTable.vue
stacytalbot: Yep - will sort

app/assets/stylesheets/application.scss
WKocur: The component styles are dependent on one another? Maybe this should be reflected in the folder structure... Or something should be moved to helper styles? I like the idea of the components being independent of each other. What do you think?

app/assets/stylesheets/styles/components/form/_select.scss
stacytalbot: sorted

app/assets/stylesheets/application.scss
stacytalbot: yea its the icons as they aren't really components but we don't have a folder for things like icons and buttons etc. yet

app/javascript/components/table/VTable.vue
stacytalbot: done

app/assets/stylesheets/application.scss
WKocur: Ok, I guess these would be resources in the nature map way of doing things. Will leave for now, but something to think about when the refresh is done.

app/assets/stylesheets/styles/components/_carousel.scss
stacytalbot: sorted

app/javascript/mixins/mixin-select-shared.js
stacytalbot: done

PR #314, opened 2019-09-13
Add regions

Levia:
## Description

* Add regions data alongside countries.
* Adjust global stats calculation for coverage and effectively managed

⚠️ This requires the following db branch: https://github.com/unepwcmc/protectedplanet-db/pull/21
⚠️ So a `bundle exec rake db:migrate` is required

## TODO

At the moment the regional stats are not 100% correct as we would need to get confirmation of the current calculations.


lib/modules/stats/country_statistics_api.rb
```diff
@@ -88,14 +97,18 @@ def get_global_stats(endpoint=nil)
...
-
-      _sum = data.inject(0) do |sum, x|
-        sum + (x[field] ? x[field] : 0)
+      attribute = ATTRIBUTES[endpoint.to_sym][:attribute]
+
+      attr_area_sum = total_area_sum = 0
+      data.map do |x|
+        attr_area = (x[attribute] ? x[attribute] : 0)
+        total_area = x[COUNTRY_AREA_ATTRIBUTE] || 0
+        attr_area_sum += (attr_area > 0 && total_area > 0) ? (attr_area * total_area / 100) : 0
```
lucacug: could either `attr_area` or `total_area` be negative?? If this is the case the check is correct, but if they can only have `0` or positive values I think you are safe even without this ternary operator

lib/modules/stats/country_statistics_api.rb
```diff
@@ -88,14 +97,18 @@ def get_global_stats(endpoint=nil)
...
       chart_json = Aichi11Target::DEFAULT_CHART_JSON.dup
-      field = ENDPOINTS[endpoint.to_sym][:field]
-
-      _sum = data.inject(0) do |sum, x|
-        sum + (x[field] ? x[field] : 0)
+      attribute = ATTRIBUTES[endpoint.to_sym][:attribute]
+
+      attr_area_sum = total_area_sum = 0
+      data.map do |x|
+        attr_area = (x[attribute] ? x[attribute] : 0)
```
lucacug: couldn't you use something like `x[attribute] || 0` in this case??

lib/modules/stats/country_statistics_api.rb
```diff
@@ -88,14 +97,18 @@ def get_global_stats(endpoint=nil)
...
+        total_area = x[COUNTRY_AREA_ATTRIBUTE] || 0
+        attr_area_sum += (attr_area > 0 && total_area > 0) ? (attr_area * total_area / 100) : 0
+        total_area_sum += total_area
       end
-      value = (_sum / data.length).round(2)
+      value = (attr_area_sum / total_area_sum * 100).round(2)
```
lucacug: I would check if the `total_area_sum` is `0` at this point otherwise I think it will complain about a division by zero

config/initializers/migration_helpers.rb
```diff
@@ -0,0 +1,9 @@
+ActiveRecord::Migration[5.0].class_eval do
+  def view_sql(timestamp, view)
+    File.read(Rails.root.join("db/views/#{view}/#{timestamp}.sql"))
+  end
+
+  def function_sql(timestamp, function)
```
lucacug: Could you explain why this is needed?? I don't find any reference in the code

lucacug:
Happy with this on the BE side of things now

PR #318, opened 2019-09-20
T11 search backend

Levia:
## Description

This is to hook up the frontend and backend together for the search to work.

⚠️ If you have already migrated your db using [this PR](https://github.com/unepwcmc/protectedplanet-db/pull/21) you would probably need to rollback (just one rollback is enough) and migrate again.
⚠️ The base branch will probably need to change when that will be merged


Levia:
@stacytalbot Ah, didn't know that. I'll provide a fix asap

lucacug:
Looks good to me. Happy to merge this!! Is the base branch the correct one??

PR #321, opened 2019-09-24
T11 adjust global stats

Levia:
## Description

A small fix to include ABNJ (Area Beyond National Jurisdiction) and ATA (Antarctica) in the global stats calculations.
Statistics objects with `nil` country_id are referring to those particular areas.

Also make sure to calculate global Pame stats correctly.


lucacug:
looks good

PR #323, opened 2019-09-26
Add default ordering by region_id, type and name

Levia:
## Description

Add default order for table data.
The order goes as follows:

1. region_id (to group a region with the related countries)
2. type (regions first, then countries)
3. name

## Notes

[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/pp-target-11-dashboard/tickets/40)

This now includes some more work that has been merged into secondary branches.

### Latest update:
Needs to be linked up to [this pp-db PR](https://github.com/unepwcmc/protectedplanet-db/pull/24)


app/serializers/aichi11_target_dashboard_serializer.rb
```diff
@@ -83,20 +83,25 @@ def serialize_options
 
   def sorted
     obj_type = @search_type == 'id' ? "AND obj_type = 'country'" : ''
+    default_order = 'ORDER BY region_id, obj_type DESC, name'
```
lucacug: Should this be the region name rather than `region_id`?? Just because I'm getting `Asia` before `Africa` in the default ordering

Levia: Thanks @lucacug!
I've fixed it now. You should link this up to [this pp-db PR](https://github.com/unepwcmc/protectedplanet-db/pull/24) now though (or merge that one first)

PR #325, opened 2019-09-26
Fix/docker and webpacker

drewpotter:
This is the initial work to fix docker webpacker support.


drewpotter:
I have spent a little bit of time this weekend thinking about and trying to fix the hot reloading for webpacker; I have pushed some changes again which will require:
```
docker-compose up --build
```
To make sure you get the ENV vars correctly in docker you can try the following command to check:
```
docker-compose run web /bin/bash -l -c 'echo "$NODE_ENV"'
```
You should see `development` is eventually outputted by this command. Also the same for this:
```
docker-compose run web /bin/bash -l -c 'echo "$RAILS_ENV"'
```

PR #326, opened 2019-09-26
Fix wdpa import  after upgrade aws-sdk gem from v1

lucacug:
This is to fix the WDPA import process after the overall upgrade of the code.
I have tested this on the new production server only and seems working fine but please double check if you notice something odd http://pp.new-web.pp-production.linode.protectedplanet.net/

:warning:  Be aware that this includes the branch `new_production_deploy` where the gem activerecord-postgis-adapter has been updated to fix the issue with the search by country.
Furthermore, the capistrano deploy files point to the new server (staging and production).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ gem 'rails', '5.0.5'
 gem 'webpacker', '~> 4.0.2'
 
 gem 'pg', '~> 0.21'
-gem 'activerecord-postgis-adapter', '~> 4.0.0'
+gem 'activerecord-postgis-adapter', '4.1.0'
 gem 'dbf', '~> 2.0.7'
 #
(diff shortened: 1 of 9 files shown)


lucacug:
:warning:  Tests are now passing, but I noticed there is this module `lib/modules/countries_geometry_Importer.rb` (not used in the import process) that still uses old `aws-sdk` gem methods and needs to be updated as well.
I'll create a new ticket on codebase for this as it will not be included in this PR.

Levia:
Everything works well for me!
The only note I would like to make is that the new version of the aws gem is installing a lot of dependencies which we probably don't need. So it would be good to tackle this at some point as the main changes involving that have already been merged.

Will merge!

PR #327, opened 2019-09-27
Tickets/57 coverage stats

benkt2:
Fix for ticket #57
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/57

### DIFF
--- a/app/presenters/statistic_presenter.rb
+++ b/app/presenters/statistic_presenter.rb
@@ -24,7 +24,15 @@ class StatisticPresenter
   end
 
   def percentage_total_pa_cover
-    percentage_pa_land_cover + percentage_pa_marine_cover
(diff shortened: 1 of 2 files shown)


app/presenters/statistic_presenter.rb
```diff
@@ -24,7 +24,15 @@ def geometry_ratio
   end
 
   def percentage_total_pa_cover
-    percentage_pa_land_cover + percentage_pa_marine_cover
+    perc = ((pa_land_area + pa_marine_area).to_f / ( land_area + marine_area ) * 100).round(2)
```
drewpotter: @benkt2 I am just wondering if we would need or benefit from a `rescue nil` - however if the data has good integrity we should never have a divide by zero anyway.

benkt2:
Was a random timing error in travis, I've seen this before which is why the integration tests sleep a bit after seeding/fixturing/factorybotting and before testing

PR #329, opened 2019-10-02
Fix sorting for coverage and effectively_managed

Levia:
## Description

Fix sorting not working for coverage and effectively managed.
This was happening because the wrong key was being passed to the hash of each record to fetch the values.

### DIFF
--- a/app/serializers/aichi11_target_dashboard_serializer.rb
+++ b/app/serializers/aichi11_target_dashboard_serializer.rb
@@ -89,7 +89,10 @@ class Aichi11TargetDashboardSerializer < CountrySerializer
     _data = ActiveRecord::Base.connection.execute(query)
 
     if @params[:sort_by].present?
-      _data = _data.sort_by { |d| d[sort_by] }
+      sort_terms = sort_by.split('+')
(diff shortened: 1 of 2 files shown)


app/serializers/aichi11_target_dashboard_serializer.rb
```diff
@@ -89,7 +89,10 @@ def sorted
     _data = ActiveRecord::Base.connection.execute(query)
 
     if @params[:sort_by].present?
-      _data = _data.sort_by { |d| d[sort_by] }
+      sort_terms = sort_by.split('+')
+      _data = _data.sort_by do |d|
+        sort_terms.inject(0) { |sum, x| sum + (d[x] || 0) }
+      end
```
lucacug: I'm getting this error now while trying to access the dashboard, I think because of the `params[:sort_by]` being present by default with value `name`
Screenshot from 2019-10-02 14-32-24

Levia: That should be fixed now, thanks @lucacug !

lucacug:
happy now, merging....

PR #334, opened 2019-10-08
Add social metadata

stacytalbot:
Add social media metadata - the content_for way of working for the marine page will be changed during the refresh so that it uses content from a yml file instead

https://unep-wcmc.codebasehq.com/projects/pp-target-11-dashboard/tickets/42


app/helpers/application_helper.rb
```diff
@@ -68,6 +68,16 @@ def encoded_page_url
...
   """
+
+  def yml_key
+    case controller_name
+    when 'target_dashboard'
+      'thematic_area.target_11_dashboard'
+    else 
+      nil
+    end
+  end
```
lucacug: While this is working fine, I would refactor using the ternary operator rather than a `case` statement, so something like:
`controller_name = 'target_dashboard' ? 'thematic_area.target_11_dashboard' : nil`

config/deploy/staging.rb
```diff
@@ -1,5 +1,5 @@
 set :stage, :staging
-set :branch, "t11-tooltips"
+set :branch, "t11-tracking"
```
lucacug: not a big deal, but I would reset this to `develop` here

stacytalbot: Yep sure i pushed this by mistake

app/helpers/application_helper.rb
stacytalbot: this is going to have a lot more than just target_dashboard in there when we do the refresh so just thought i was start it as it will need to go on.

lucacug: Oh, I see. Let's keep it as it is then.

lucacug:
cool, happy now

PR #342, opened 2019-10-23
Feature/acp regions

Levia:
## Description

This is to amend the representation of biopama countries and replace the `is_biopama` bool with an `acp_region` string to also represent the ACP region countries belong to.
This means that countries with `acp_region` equal to `nil` are not ACP countries.

This PR specifically is about updating the CSV source and the importer.
Changes made on the db can be found on the [related pp-db PR](https://github.com/unepwcmc/protectedplanet-db/pull/26).

## Notes

* [Related PP-API PR](https://github.com/unepwcmc/protectedplanet-api/pull/10)
* Run importer: `Wdpa::BiopamaCountriesImporter.import`
* Regions for ACP countries have been provided by JRC as those names do not match with already existing regions. This is also the reason why these regions have been currently added as a new field in the `countries` table. Another potential approach would have been doing a Single Table Inheritance with the region model so that, in the future, we could add different regions for different mappings. However, that requires important changes within the code which would need to be properly scheduled in.


lib/modules/wdpa/biopama_countries_importer.rb
```diff
@@ -1,18 +1,19 @@
 module Wdpa::BiopamaCountriesImporter
-  BIOPAMA_COUNTRIES_CSV = "#{Rails.root}/lib/data/seeds/biopama_countries_iso_codes.csv"
+  BIOPAMA_COUNTRIES_CSV = "#{Rails.root}/lib/data/seeds/biopama_countries.csv"
   extend self
 
   def import
     ActiveRecord::Base.transaction do
       csv = CSV.read(BIOPAMA_COUNTRIES_CSV)
       csv.shift # remove headers
```
lucacug: I think these two lines can be removed now you are using the `foreach` at line 10

Levia: Ah, I forgot about those. Fixed now. Thanks, @lucacug !

lucacug:
Happy now!!

Levia:
@stacytalbot We were waiting for JRC to approve this before merging and going on production, but I think they never came back to us (or at least to me). Might be worth double checking internally with the rest of the team.

PR #358, opened 2020-01-24
Refresh nav

stacytalbot:
:warning: Merge #357 first

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/49
- Also check `def get_nav_primary` in `application_helper.rb`


app/helpers/application_helper.rb
```diff
@@ -24,6 +24,10 @@ def yml_key
     end
   end
 
+  def active_nav_item?(test_path)
```
Levia: I feel like the argument variable should be called differently to be clearer on what that is

stacytalbot: @Levia changed to `current_path`

PR #359, opened 2020-01-27
Refresh drop legacy pas

Levia:
## Description

* Remove legacy pas related code
* Adjust tests
* Update docs

## Notes

Related [db PR](https://github.com/unepwcmc/protectedplanet-db/pull/30).
Related [codebase ticket](https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/67).


test/controllers/sites_controller_test.rb
```diff
@@ -1,16 +1,14 @@
...
-    FactoryGirl.create(:legacy_protected_area, wdpa_id: wdpa_id, slug: slug)
 
-    get :show, params: {id: slug}
+    get :show, params: {id: wdpa_id}
 
-    assert_redirected_to "/#{protected_area.slug}"
+    assert_redirected_to protected_area_path(id: protected_area.slug)
   end
 
   test 'GET :show redirects to the homepage if no match legacy PA
```
stacytalbot: Can this test be deleted?

Levia: Yeah you are right. I got rid of everything related to sites routes etc. as this basically turns out to be another way of accessing the same resource (PAs) now that we removed legacy PAs.

Thanks!

lucacug:
@stacytalbot @Levia Happy with this, the only thing I noticed is the build is not passing for an error in one of the tests. Do you want to fix this here, or do you want me to merge this anyway?

Levia:
Thanks @lucacug !

The errors should be related to something else.
We have decided to take care of the tests later as there will still be a lot of refactoring going on, so it would probably make more sense to tackle those when we have finished that.

PR #360, opened 2020-01-29
Refresh footer

stacytalbot:
⚠️ Merge #358 first

- Add all footer content and style responsively

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/38


app/helpers/application_helper.rb
```diff
@@ -334,4 +328,25 @@ def get_thematical_areas
       "cards": thematical_page.children.published
     }
   end
+
+  def get_footer_links
+    def make_links slug_array
+      array = []
+      
+      slug_array.each { |slug|
```
Levia: Two quick comments here:

1. When it's about a multi-line block, it would be better to use `do ... end` instead of curly brackets for better readability.
2. You can probably use `map` instead of `each` so that the last thing you do inside the block gets returned as an element of the array. In this way you won't need to do an extra assignment to another array but just returning the hash for each element

app/helpers/application_helper.rb
```diff
@@ -334,4 +328,25 @@ def get_thematical_areas
       "cards": thematical_page.children.published
     }
   end
+
+  def get_footer_links
+    def make_links slug_array
```
Levia: Also please do not define nested methods.
If you know a method is only going to be used from other methods, define this method as private.

In ruby you can start a list of private methods by just using `private` before those definitions, like:

```
  private

  def my_method
  end

  def another_method
   end
```

stacytalbot: Moved nested footer and primary nav methods to private ones

app/helpers/application_helper.rb
stacytalbot: Swapped out curly braces for `do ... end` and used map instead of each

PR #361, opened 2020-01-29
Refresh heros

stacytalbot:
⚠️ Merge #360 first
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/40
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/41


app/views/layouts/partials/_hero-basic.html.erb
```diff
@@ -1,5 +1,20 @@
-<header class="hero--basic">
+<% classes = '' if local_assigns[:classes].nil? %>
```
Levia: Could the assignments on the html.erb files be transformed into helper functions instead please?

PR #368, opened 2020-03-09
Allow to split data into multiple shapefiles

Levia:
## Description

It has been reported that the shapefile download generates a unique large shapefile which is diffcult or impossible to work with (probably due to the 2GB limitation for shapefiles).

This PR has got some adjustments that would split the data into 2 shapefiles by default, but more generally allows to specify the number of shapefiles desired when exporting.

## Notes

Needs to be tested on staging properly.

[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/112)

### DIFF
--- a/lib/modules/download/generators/shapefile.rb
+++ b/lib/modules/download/generators/shapefile.rb
@@ -13,7 +13,7 @@ class Download::Generators::Shapefile < Download::Generators::Base
     }
   }
 
-  def initialize zip_path, wdpa_ids, number_of_pieces=2
+  def initialize zip_path, wdpa_ids, number_of_pieces=3
     @path = File.dirname(zip_path)
     @filename = File.basename(zip_path, File.extname(zip_path))
     @wdpa_ids = wdpa_ids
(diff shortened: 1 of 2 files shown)


lucacug:
Had a look at this and the code looks good. Testing this locally I got odd results though: 
the shape file was split into two, but with a huge difference in size (~2.5 GB vs ~200 MB). 
I suggest testing this on staging, as this may be due to my local db.

Levia:
@lucacug That's probably because I'm ordering the records by descending area size:
```'ORDER BY \""GIS_AREA"\" DESC'```

Which I guess makes the biggest polygons to be exported in the first shapefile.

Perhaps we could do some testing without that bit so that areas are allocated randomly across different shapefiles. I didn't envision the first shapefile could get that large that way, which should possibly never get larger than 2GB.

Otherwise it's either splitting in 3 parts or try some more cleaver sorting

Levia:
@lucacug I've changed the order to be by ascending WDPAID.
It seems to be better this way.

lucacug:
@Levia  Thanks it is better now, I got two zip files of pretty much the same size (1.2GB vs 1.9GB). Could I suggest we split this into more than 2 (maybe 4 ?) distinct shapefiles, as once unzipped one of them it is still bigger than 2GB (~2.6)

Levia:
@lucacug Thanks! I've pushed the related changes

PR #370, opened 2020-03-13
Refresh region

stacytalbot:
- Start to create components for site, country and region pages.


Levia:
@stacytalbot I have pushed some commits related to refactoring and adding in some missing values.

I'm generally happy with this PR now, although I was wondering if do you think it might be worth simplifying the `show` views a bit adding some partials?
It feels like there's quite a lot of stuff in just one html file.

Levia:
Merging, other fixes/refactoring to be done later

PR #371, opened 2020-03-26
Refresh marine

stacytalbot:
http://localhost:3000/en/thematical-areas/marine-protected-areas
- Please see tickets for details


Levia:
@stacytalbot Things looks fine to me.
However, it would be good if we could take this opportunity for properly adjust the backend related to the data, instead of having things hardcoded in the controllers.
Probably something for another PR but I thought it would have been good mentioning it.

Is the do not merge label there because this needs more adjustments on the FE/Design side of things?

PR #376, opened 2020-04-01
Refresh marine line

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/22


app/controllers/marine_controller.rb
```diff
@@ -153,236 +153,80 @@ def marine_statistics
   end
 
   def growth
+    # @protectedAreasGrowth = [
+    #   {
+    #     id: 'a',
+    #     datapoints: [{x:0,y:0},{x:1,y:1},{x:2,y:2},{x:3,y:3},{x:4,y:4},{x:5,y:5},{x:6,y:6},{x:7,y:7}]
+    #   }
+    # ].to_json
```
Levia, line 161: Can this be removed completely?

stacytalbot: yep will do that now

PR #378, opened 2020-04-02
Make CTAs as CMS components

Levia:
## Description

Make CTAs as CMS custom component.

Tickets:

- https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/30
- https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/31

Related DB PR: https://github.com/unepwcmc/protectedplanet-db/pull/33


app/views/admin/ctas/index.html.haml
```diff
@@ -0,0 +1,27 @@
...
+      %td.wrap= cta.title
+      %td.wrap= cta.summary
+      %td= cta.url
+      %td= cta.updated.to_s
+      %td
+        .btn-group.btn-group-sm
+          = link_to 'Edit', edit_admin_cta_path(cta), :class => 'btn btn-default'
+          = link_to 'Delete', admin_cta_path(cta), :method => :delete, :data => { :confirm => 'Are you sure?' }, :class => 'btn btn-danger'
+
+= will_paginate @ctas, :theme => 'comfy'
```
stacytalbot: @Levia didn't know you were a fan of haml?

Levia: @stacytalbot I'm not :) 
I just kept it consistent with the other CMS custom component, but might think about changing it if the carousel slides will go away.

Levia:
@stacytalbot I've added some quick tests.
I think we could merge this and the related db branch if you are happy. Perhaps the only thing missing is to link up the css class?

app/views/admin/call_to_actions/index.html.erb
```diff
@@ -0,0 +1,33 @@
+<div class="page-header">
+  <%= link_to 'New Call To Action', new_admin_call_to_action_path, class: 'btn btn-secondary float-right', title: 'Create new call to action' %>
+  <h2>Call To Actions</h2>
+</div>
+
+<!-- = will_paginate @ctas, :theme => 'comfy' -->
```
stacytalbot, line 6: @Levia I've put this back into html so that it is easy for us to update and tidied up the page a bit. I couldn't work out what this should be though? Also not sure we need pagination for this page as there shouldn't be loads of CTAs in the site so we could just remove it.

Levia: Sounds good. And yes, I'm happy for that to be removed.
I'll do that myself and also convert the other views into full html

PR #383, opened 2020-04-16
Fix countries assessments count mismatch

Levia:
## Description

Add country association regardless of wdpaid being present.
[Ticket](https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/116)

## Notes

`Wdpa::PameImporter.import` and then `ImportTools.statistics_monthly_import` are required to apply the fix to the db.


Levia:
Merging after approval from client

PR #385, opened 2020-04-17
Fixed wrong download by making token depend on whole of filter hash

benkt2:
Tests for same


lib/modules/download/requesters/search.rb
```diff
@@ -16,19 +16,19 @@ def domain
...
+
 
   def identifier
     token
   end
 
   def token
     @token ||= begin
-      filters_dump = Marshal.dump filters.keys.sort
+      filters_dump = Marshal.dump filters.sort.to_json
```
Levia, line 27: This works well and it seems to fix the issue!
However, I'm wondering if:

1. We could fix the `sort` deprecation warning and to something like `filters.to_hash.sort.to_json` to fix it.
2. Just use values instead, like `filters.values.sort`. I don't think there will be values which are the same across the different filters, but we could just go for option 1 to be extra safe.

lib/modules/download/requesters/search.rb
```diff
@@ -16,19 +16,19 @@ def domain
...
       Digest::SHA256.hexdigest(@search_term.to_s + filters_dump)
     end
   end
-
+  private
```
Levia, line 31: it might be good to add a space before and after private to improve readability.

test/unit/workers/download_workers/search_test.rb
```diff
@@ -24,6 +25,74 @@ def setup
...
+555556064,
+2325,
+555556065,
+555556067,
+555556066,
+2327,
+317195,
+5086,
+20171,
+902264]
```
Levia, line 78: I would indent this to improve readability.

PR #396, opened 2020-06-11
Refresh design tweaks

stacytalbot:
1. Pull the latest on the refresh branch in /db
2. import cms seeds
3. Amends are across the site and will be checked by Charlotte in the browser

### DIFF
--- a/app/views/partials/download/_download.html.erb
+++ b/app/views/partials/download/_download.html.erb
@@ -1,3 +1,9 @@
-<download text="<%= t('global.button.download') %>">
+<download 
+  <% if local_assigns.has_key? :classes %>
+    class="<%= classes %>"
+  <% end %>
+
+  text="<%= t('global.button.download') %>"
+>
   <%= download_dropdown("marine", "general", [:csv, :shp, :pdf]) %>
 </download>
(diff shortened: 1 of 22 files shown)


app/views/partials/download/_download.html.erb
```diff
@@ -1,3 +1,9 @@
-<download text="<%= t('global.button.download') %>">
+<download 
+  <% if local_assigns.has_key? :classes %>
```
Levia, line 2: Is this strictly necessary? If classes is not defined you should get `nil` by default which should then result by an empty string think

stacytalbot: So you're saying add a class attribute regardless as to whether it has and classes

Levia: Yeah I was just wondering if that would cause issues.
But if it's better not to have the class attribute if classes are empty or if HTML doesn't manage this automatically then happy to leave that as is

stacytalbot: No it wouldn't cause issues.

Trying to print out `classes` would though if it didn't exist

PR #398, opened 2020-06-17
Refresh carousel

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/20


defaye:
Could add a validator to the props e.g.
```
validator: options => ["small", "medium", "large"].every(attr => attr in options && options[attr] instanceof FlickityOptions),
```
As discussed in Slack, responsiveness could possibly be fixed by adding an integer data key on the component e.g. <flickity :key="refreshKey" /> where that key is incremented on breakpoint events after the flickityOptions has been changed

PR #400, opened 2020-06-22
Pame integration

benkt2:
Help! Front-end person please fix this mess!


Levia:
@benkt2 , @stacytalbot I'm guessing the about and methodology pages are something to worry about later?

Also I'm about to resolve conflicts with refresh as well, I might need your input for some of those

app/models/pame_evaluation.rb
```diff
@@ -1,7 +1,350 @@
...
+
+SELECT_STATEMENT = [
+    'DISTINCT pame_evaluations.id AS assessment_id', 'pame_evaluations.metadata_id AS metadata_id',
+    'pame_evaluations.url AS url', 'pame_evaluations.year AS year',
+    'pame_evaluations.methodology AS methodology', 'pame_evaluations.wdpa_id AS wdpa_id',
+    'pame_sources.data_title AS source_data_title', 'pame_sources.resp_party AS source_resp_party',
+    'pame_sources.year AS source_year', 'pame_sources.language AS source_language',
+    'pame_evaluations.protected_area_id AS protected_area_id', 'protected_areas.name AS protected_area_name',
+    'designations.name AS designation', 'ARRAY_TO_STRING(ARRAY_AGG(countries.iso_3),\';\') AS countries'
+].freeze
```
Levia, line 230: It looks like this is not used anywhere?

app/models/pame_evaluation.rb
```diff
@@ -1,7 +1,350 @@
...
+
+GROUP_BY = "pame_evaluations.id, protected_areas.wdpa_id, protected_areas.name, designation, pame_sources.data_title,
+            pame_sources.resp_party, pame_sources.year, pame_sources.language".freeze
```
Levia, line 233: Same for this

app/models/pame_evaluation.rb
```diff
@@ -1,7 +1,350 @@
...
+
+  def self.generate_csv(where_statement, restricted_where_statement)
+    where_statement = where_statement.empty? ? '' : "WHERE #{where_statement}"
+    restricted_where_statement = restricted_where_statement.empty? ? '' : "WHERE #{restricted_where_statement}"
+    query = <<-SQL
```
Levia, line 238: Could we make this a separate private method at least? So to clean the code a little bit

test/unit/wdpa/pame_importer_test.rb
```diff
@@ -16,9 +16,9 @@ class TestPameImporter < ActiveSupport::TestCase
     assert_equal 9, pame_evaluations.count
   end
 
-  test "#import pame evaluations with hidden evaluation" do
-
-    PAME_EVALUATIONS = "#{Rails.root}/lib/data/seeds/test_pame_data_hidden.csv".freeze
+#  test "#import pame evaluations with hidden evaluation" do
+  def badger
```
Levia, line 20: What is this about exactly?

app/models/pame_evaluation.rb
```diff
@@ -1,7 +1,350 @@
...
+  validates :methodology, :year, :metadata_id, presence: true
+
+  # config for loadable mixin
+  ignore_column :designation
+
+  import_by protected_area: :wdpa_id
+  import_by country: :iso_3
+  
+
+  TABLE_ATTRIBUTES = [
```
Levia, line 20: Could we add `.freeze` to constants to make sure Rails throws an error if anything attemps to modify them?

app/models/pame_evaluation.rb
```diff
@@ -1,7 +1,350 @@
...
+      title: 'Link to assessment',
+      field: 'url'
+    },
+    {
+      title: 'Metadata ID',
+      field: 'metadata_id'
+    }
+  ]
+
+  SORTING = ['']
```
Levia, line 60: This seems to be unused.

app/models/pame_evaluation.rb
```diff
@@ -1,7 +1,350 @@
...
   has_and_belongs_to_many :countries
 
-  validates :methodology, :year, :metadata_id, :url, presence: true
+  validates :methodology, :year, :metadata_id, presence: true
+
+  # config for loadable mixin
+  ignore_column :designation
+
+  import_by protected_area: :wdpa_id
+  import_by country: :iso_3
```
Levia, line 17: Running the importer again I found it creates countries objects and without iso_3 (which by the way makes line 197 of this file to fail - adding `.compact` would fix it).

I'm not familiar with the Loadable gem but I wonder if this import_by line should be removed. What do you think? @benkt2

Levia: @benkt2 To add more details. I tried importing with and without this line but for me it seems to link all evaluations to the same WDPAID and Country. Removing this line seems at least to prevent generating more countries.
Does this happen for you as well?

test/unit/wdpa/pame_importer_test.rb
benkt2: badly skipped test - I have fixed test and extended the expectations to check the three restricted cases here

app/models/pame_evaluation.rb
benkt2: Yep, needed to bump wcmc_components to 0.0.5 and change this to 
import_by countries: :iso_3
Have added checks to tests to make sure evaluations get hooked up to existing country and don't create a spurious new one

app/models/pame_evaluation.rb
benkt2: correct - deleted

app/models/pame_evaluation.rb
benkt2: correct - deleted

Levia:
@stacytalbot Is this ready to be reviewed again? Also, could you please fix the conflicts when you get the chance?

Levia:
@benkt2 Running the importer still assigns all evaluations to the same PA.
Doesn't this happen to you?

I ran bundle install as you have upgraded the wcmc gem but no luck still

app/javascript/components/pame/Row.vue
```diff
@@ -0,0 +1,87 @@
...
+
+    computed: {
+      projectTitle () {
+        return this.trim(this.item.title)
+      }
+    },
+
+    methods: {
+      wdpaUrl (wdpaId) {
+        return `https://protectedplanet.net/${wdpaId}`
```
Levia, line 46: @stacytalbot @benkt2 This should better be a URL constructed based on the environment. Probably something the backend che pass with the rest of the data

stacytalbot: @Levia do you want to update the backend and I can do the front?

Levia: Yeah happy to do that

Levia: @stacytalbot I've done both FE and BE now

Levia:
Going to merge this. I've created other tickets to possibly fix the remaining bits. Ticktes ids: 149 and 150

PR #419, opened 2020-07-09
Re-arrange VMap components to handle header at mobile BP

defaye:


app/assets/stylesheets/components/_map.scss
```diff
@@ -1,30 +1,61 @@
...
+    }
+
+    .v-map-filters {
+      @include breakpoint($medium) {
+        position: absolute;
+        width: 33.33%;
+        z-index: 1;
+
+        top: rem-calc(42);
+        left: $gutter-medium * 2;
```
WKocur, line 40: We have some conventions on the order of styles, not sure if you've seen this list before: https://github.com/unepwcmc/wiki/wiki/Styling

app/assets/stylesheets/components/maps/_v-map-header.scss
```diff
@@ -0,0 +1,20 @@
...
+  font-size: rem-calc(25);
+  font-weight: 600;
+  justify-content: space-between;
+  line-height: $header-font-line-height;
+  padding: rem-calc(14 24);
+
+  .close-button {
+    margin-left: rem-calc(12);
+    &::after {
+      content: "⨯";
```
WKocur, line 15: In the styles there should be an actual icon for this - if you click the insect icon you should be able to download the assets.

app/assets/stylesheets/components/maps/_v-map-header.scss
```diff
@@ -0,0 +1,20 @@
+.v-map-header {
+  background-color: $grey-black;
+  color: $grey-xlight;
+  display: flex;
+  font-family: $header-font;
+  font-size: rem-calc(25);
+  font-weight: 600;
```
WKocur, line 7: Is there a font weight variable (bold or semi bold) that we can use here?

app/views/partials/maps/_main.html.erb
```diff
@@ -1,10 +1,9 @@
 <div class="map--main">
+  <v-map-header title="<%= @main_map[:title] %>"></v-map-header>
+  <v-map></v-map>
   <v-map-filters
     title="<%= @main_map[:title] %>"
     :overlays="<%= @main_map[:overlays].to_json %>"
     :disclaimer="<%= @main_map[:disclaimer].to_json %>"
-  >
-    <v-map-p-a-search></v-map-p-a-search>
```
WKocur, line 7: Do you plan to add this back in?

defaye: Isn't it used via the Filters component(?)

app/assets/stylesheets/components/maps/_v-map-header.scss
defaye: Ah yes, I haven't noticed. Will use $semi-bold

app/assets/stylesheets/components/_map.scss
defaye: Will adapt!

app/views/partials/maps/_main.html.erb
WKocur: I was initially passing it in as a slot, in the hope that it could be a bit separate to the more reusable/generic map component. I don't know how feasible that separation will be though, as I'm not too sure how the search will work yet...

app/assets/stylesheets/components/_map.scss
WKocur: Cheers!

defaye:
Thanks for the link to the style guide. With regards to BEM, I will try to start using it more strictly! I was wondering if we FE folk could have a look at some alternatives for future projects like [ABEM](https://css-tricks.com/abem-useful-adaptation-bem/) though? @stacytalbot 

The problem with BEM is that you end up with (as the article points out quite well) massive class trains in the HTML, which can really clutter the HTML up. I found this frustrating when using BEM in the past and I think I naturally started using BEM in a slightly different way (like ABEM, but without the [a/m/o] prefixing). Perhaps it is something we can have a think about in the future? The author of the ABEM article does highlight some potential drawbacks of ABEM, but offers some solid advice when such edge-cases arise. I also particularly like the issue it highlights with the hyphenation of block-level class names.

I will have to refresh myself on BEM quickly, but I can adapt my styles to the BEM syntax for sure @WKocur

app/assets/stylesheets/components/maps/_v-map-header.scss
defaye: I tried using the included icon but the SVG itself (if not mistaken) has an embedded circle around the "X". Maybe I can retry using it, but I fear the icon will display too small relative to the header text because it is wrapped in a circle. That's why I opted for this instead.

app/views/partials/maps/_main.html.erb
defaye: Can I leave it out until such time we figure this out then?

WKocur: Sure - in that case I would remove it from vue.js (or application.js... I forget the file name) as well as it isn't used anywhere else. Thanks!

app/assets/stylesheets/components/maps/_v-map-header.scss
WKocur: I found one that should work - I'll send ver slack as it doesn't want to upload here.

WKocur:
Cool - sounds good. @defaye I'm happy to merge this once you've done the above. There are a few touch ups in terms of styling - margin/paddings etc. are a little out in places. If you use the inspect mode in xd, you should be able to get the exact values from the designs. Also note the disclaimer text is white in the designs.

In summary, this works great for the structure - I reckon good to merge in once you've changed the classes and ordering of styles. The touching up of css values can then be done on another branch as it should affect any other work.

Oh one thing to check: @stacytalbot which breakpoint is the change from mobile to tablet? Small? @defaye I think it might be switching between mobile and tablet at one breakpoint too high, but I'm out of the loop here so might be wrong!

PR #425, opened 2020-07-14
Map country and region

WKocur:
Adds map to country and region pages.
Zooms to region/country on load - currently slow for regions.

e.g. /country/gbr or /region/europe

### DIFF
--- a/app/assets/stylesheets/components/_map.scss
+++ b/app/assets/stylesheets/components/_map.scss
@@ -1,4 +1,5 @@
 @import "maps/v-map-baselayer-controls";
+@import "maps/v-map-disclaimer";
 @import "maps/v-map-filter";
 @import "maps/v-map-filters";
 @import "maps/v-map-header";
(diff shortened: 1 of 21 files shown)


app/assets/stylesheets/components/_map.scss
```diff
@@ -40,6 +41,20 @@
     }
   }
 
+  &--header {
+    width: 100%;
```
defaye, line 45: I'm also guilty of this, but isn't BEM for element level classes meant to be underscore-prefixed?

WKocur: This is a modifier (for a map that's in the header, as opposed to a map header). Header probably isn't the best word. I can try and think of a better one. It is quite misleading!

defaye: Ah I see what you mean, okay. If you're happy I can merge, unless you have any last minute changes?

PR #427, opened 2020-07-14
Refresh maps dropdown and search

WKocur:


app/assets/stylesheets/components/maps/_v-map-pa-search.scss
```diff
@@ -0,0 +1,37 @@
+.v-map-pa-search {
+  margin-bottom: rem-calc(12);
+
+  position: relative;
+
+  display: flex;
```
WKocur, line 6: display above position

app/assets/stylesheets/components/maps/_v-map-pa-search.scss
```diff
@@ -0,0 +1,37 @@
...
+
+  &__input {
+    background-color: rgba($black, 0);
+    border-radius: rem-calc(32);
+    border: rem-calc(1) solid $white;
+    outline: none;
+    padding-left: rem-calc(50);
+    color: $white;
+    font-size: rem-calc(18);
```
WKocur, line 15: alphabetic order

app/assets/stylesheets/components/select/_selector.scss
```diff
@@ -0,0 +1,75 @@
...
+        cursor: default;
+
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+
+        &:focus {
+            box-shadow: 0 0 0 rem-calc(2) rgba($grey-light, 0.25);
+        }
+    }
```
WKocur, line 20: I personally think it's a bit easier to read with a break line between the classes.

app/assets/stylesheets/components/select/_selector.scss
```diff
@@ -0,0 +1,75 @@
...
+        border-top: rem-calc(1) solid $grey-xlight;
+        border-bottom: rem-calc(1) solid $grey-xlight;
+        color: $white;
+
+        position: absolute;
+        top: rem-calc(52);
+        left: 0;
+        width: 100%;
+        height: rem-calc(100);
+        overflow-y: scroll;
```
WKocur, line 45: order

app/javascript/components/map/VMapFilters.vue
```diff
@@ -1,52 +1,41 @@
...
+        @change="onDropdownChange"
+      />
+      <v-map-pa-search @search="onSearch" :type="searchType" />
       <div class="v-map-filters__overlays">
-        <div
-          v-for="(overlay, index) in overlays"
-          :key="index"
-          class="v-map-filters__overlay"
-        >
+        <div v-for="(overlay, index) in overlays" :key="index" class="v-map-filters__overlay">
```
WKocur, line 13: Can we keep the broken lines for more than one attribute? I think this is in the linter.

WKocur: You should just be able to run yarn lint on this file 🤞 Single quotes in the javascript is also set in the eslint config.

app/javascript/components/map/VMapFilters.vue
```diff
@@ -58,29 +47,76 @@ export default {
...
   },
 
-  data () {
-    return {
-      show: true
+  computed: {
+    dropdownOptions() {
+      return this.searchTypes.map(type =>
+        this.convertSearchTypeToDropdownOption(type)
+      );
```
WKocur, line 87:
```
return this.searchTypes.map(this.convertSearchTypeToDropdownOption)
```
should work

app/javascript/components/map/VMapPASearch.vue
```diff
@@ -1,17 +1,88 @@
...
+      this.$refs.input.focus();
+    },
+
+    onIconClick() {
+      if (this.query) {
+        this.submitSearch();
+      } else {
+        this.focusInput();
+      }
+    },
```
WKocur, line 66: Might be worth checking if the autocomplete is meant to update automatically as you type. I suppose this depends how well it can cope with that many requests.

app/javascript/components/map/VMapPASearch.vue
```diff
@@ -1,17 +1,88 @@
...
+          this.autocompleteResults = results;
+        })
+        .catch(e => {
+          console.error(e);
+          this.autocompleteResults = [];
+        });
+    },
+
+    submitSearch() {
+      return axios.post("/search/autocomplete", {
```
WKocur, line 81: I think you may need to add a header to axios (see Stacy's axios mixin)

app/javascript/components/select/Selector.vue
```diff
@@ -0,0 +1,200 @@
...
+          label: ""
+        },
+        ...this.options.map(opt => {
+          if (typeof opt === "string") return { value: opt, label: opt };
+          return {
+            value: opt.value,
+            label: opt.label
+          };
+        })
+      ];
```
WKocur, line 101: Perhaps worth popping this mapping function into a separate method

app/javascript/components/select/Selector.vue
```diff
@@ -0,0 +1,200 @@
...
+export default {
+  model: {
+    event: "change",
+    prop: "state"
+  },
+  props: {
+    default: {
+      required: false
+    },
+    state: {
```
WKocur, line 50: state could could perhaps be renamed slightly to make it clearer that it's the selected value (if it that is what it is)

app/javascript/components/select/Selector.vue
```diff
@@ -0,0 +1,200 @@
...
+            (this.search.length === 1 ? "^" : "") + this.search,
+            "i"
+          ).test(option.label)
+        )[0];
+        if (searchResult) {
+          if (this.dropdownEnabled) {
+            // focus if you have the options in front of you to choose from
+            this.$refs.options[
+              this.computedOptions.slice(1).indexOf(searchResult)
+            ].focus();
```
WKocur, line 159: How well does this work if the dropdown has a scroll?

app/javascript/components/map/VMapFilters.vue
defaye: What's the reason for single quotes in the lint config given that JS doesn't differentiate between the two types?

app/javascript/components/map/VMapPASearch.vue
defaye: I was planning on implementing this actually!

app/javascript/components/map/VMapFilters.vue
WKocur: Consistency and to avoid merge conflicts. No strong reason for single vs double I don't think.

app/javascript/components/map/VMapPASearch.vue
defaye: Can we not just bind axios on the window object or on the Vue prototype e.g. Vue.prototype.$http = axios so we can use it from any component as this.$http without the need to re-import? We can set all the constraints on axios via something like axios.defaults.headers - it then saves us having to do this every time

app/javascript/components/select/Selector.vue
defaye: Works fine as far as I can see, are any particular concerns you were thinking of?

app/javascript/components/map/VMapPASearch.vue
WKocur: This makes sense to me, would be good if you could do that. If you do this, can you leave the map request helpers as is? They use different axios config.

app/javascript/components/select/Selector.vue
WKocur: The page can sometimes jump around weirdly when focusing an object, rather than just scrolling within the options container.

app/javascript/components/map/VMapPASearch.vue
WKocur: Cool! Check with Ben or Ferdi first probably just to make sure the backend will support this fine. Also there is already a search comp in the site. Have you looked to see if any of that is reusable?

app/javascript/components/map/VMapPASearch.vue
defaye: If there's a slightly different config, I see what you mean. I can implement this.$http more generally if we agree, I'll have a look at the axios mixin, I'm not yet clear on what's needed

app/javascript/components/select/Selector.vue
defaye: Ah I see, I don't think it matters because it's in an overflow-y scroll, but it is difficult to prevent body scroll if using the arrow keys to tab between the options (I did think about looking into it), but decided against this because of time

WKocur: Shouldn't be a problem here, as it's a short list anyway, but worth thinking about if it gets longer in the future. I think there's something in O+ habitats that scrolls acceptably (if not amazingly!)

app/javascript/components/map/VMapFilters.vue
defaye: Ah I see. I've just setup ESLint on VSCode so this is going to do a lot of the work for me

PR #429, opened 2020-07-15
Move urls to backend

WKocur:
This means all data about map service urls is kept in the same place `map_helper.rb` in the backend.

Can be tested on a page such as 'country/FRA' or 'region/EU' (the map takes a while to load on the latter - need to chat about this)


config/database.yml
```diff
@@ -14,7 +14,7 @@ default: &default
 
 development:
   <<: *default
-  database: protectedplanet-db
+  database: protectedplanet-db-oecm
```
defaye: Should this by loaded from the environment file?

```
development:
  <<: *default
  database: <%= ENV['POSTGRES_DBNAME'] %>
```

PR #430, opened 2020-07-15
Refresh maps dropdown and search v2

defaye:
image
image
image
image


defaye:
I've made a first wave of changes @WKocur 

I don't have time yet to look into the tab out behaviour. The issue is that I need to register a keyup; key===tab event on the document itself and would currently be coded in a very similar way to the `onDocumentClick` handler in Selector.vue.

Todo: thinner scrollbar in autocomplete and extra autocomplete functionality and IE compatibility checks/fixes

app/javascript/components/autocomplete/Autocomplete.vue
```diff
@@ -61,11 +62,7 @@ export default {
      * @return void
      */
     eventHandler(document, 'keyup', function (e) {
-      if (e.key.toLowerCase() === 'tab') {
-        if (!this.$el.contains(document.activeElement)) {
-          this.resetAutocompleteResults()
-        }
-      }
+      if (e.key.toLowerCase() === 'tab') this.onFocusCheck()
```
WKocur, line 65: FYI we do have mixin-popup-close, focus-helpers and keyboard-helpers which aim to centralise some of this stuff. Could this either be incorporated into them or them into this?

WKocur: (They are in the PP repo already)

defaye: Sure. I am sure the functionality of what I have may overlap though.

I could keep my event handler mixin in case there are other events not to do with esc/tab etc that come up in the future though as it is really simple to understand when you can use it.

```
    eventHandler(document, 'keyup', function (e) {
      if (e.key.toLowerCase() === 'tab') this.onFocusCheck()
    }),
```
Specify the element, the event to listen for, and a handler, the rest is up to you.

WKocur: I think this is useful. The current event mixins (focus capture and popup close) should use this one, if this is how we're going to handle events. I think it's best to keep our approaches consistent.

Will this return true for shift + tab as well as just tab? For keycodes, I think you have to specify whether shift is or isn't present. If you do, I would recommend updating focus helpers to use e.key instead of e.keycode and then using the isTabForward/isTabBackward functions here. Unless of course you want to listen to both tab forward and backward.

WKocur:
@defaye Is the select working in that you can search only by PAs, countries or regions? I've only seen PA results so far, but haven't tested much. It seems quite slow - is it for you too?

defaye:
> @defaye Is the select working in that you can search only by PAs, countries or regions? I've only seen PA results so far, but haven't tested much. It seems quite slow - is it for you too?

I did notice this as well. I think I may need to check the params I'm sending but it should be capable of this.
It does have a little bit of debouncing applied, which I could reduce the timeout for - it is current 100ms.

PR #433, opened 2020-07-15
Ticket 53

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/53

Created a Vue component allowing users to select and display the various graphs and maps for each protected area, as well as a custom dropdown component.


WKocur:
@stanleypliu Which page is this - do I need to add something in the cms?

WKocur:
Screenshot 2020-07-16 at 13 20 18
I'm also getting this error - not sure if this is was introduced on this branch or before

WKocur:
Screenshot 2020-07-16 at 13 27 26
Is it possible to change the field type here so we have a larger text area and perhaps a wysiwyg editor? @stacytalbot were there any requirements around this? Might be a wider cms thing to separate into another ticket.

stanleypliu:
> @stanleypliu Which page is this - do I need to add something in the cms?

Delete the homepage and re-import the seeds via the rake task so that the import works correctly if you can't view a particular page. It's on the equity and results page.

app/assets/stylesheets/_settings.scss
```diff
@@ -10,8 +10,10 @@ $white: #ffffff;
 $black: #000000;
 
 $grey-xlight: #f4f4f4;
+$grey-med-light: #ededed;
 $grey-light: #c8c8c8;
 $grey: #919191;
+$grey-med-dark: #707070;
```
stacytalbot, line 16: Charlotte shouldn't be adding in anymore greys at this stage - can you just use the closest of the existing ones and delete these 2 new ones.

app/assets/stylesheets/base/_buttons.scss
```diff
@@ -208,4 +208,8 @@ $padding-medium: rem-calc(27);
   &--outline-black { @include button-outline($black); }
   &--outline-white { @include button-outline($white); }
   &--primary { @include button-block; }
+  &--view-pa {
```
stacytalbot, line 211: Here can you assign the .button--primary class to the button so that it looks as it should. Then any additional styles to that button can be added with another class such as .select__button or .select__panel-button

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,102 @@
...
+    &__options {
+        display: none;
+
+        &--active {
+            height: rem-calc(500);
+            overflow: scroll;
+            position: absolute;
+            display: block;
+            top: 100%;
+            left: 0;
```
stacytalbot, line 69: Can you just tidy up the ordering of some of the styles in this file? i.e.

&--active {
            border: 1px solid $grey-black;
            border-top: 0;
            background: $white;
            height: rem-calc(500);
            overflow: scroll;
            
            display: block;
            position: absolute;
            top: 100%;
            right: 0;
            left: 0;
            z-index: 2;

stanleypliu:
> @stanleypliu Responsive styling will need to be added: If you click the squares icon in the top left hand corner of XD it takes you back to see all the different page designs. There is style for both mobile and tablet.

My bad, didn't realise there were further designs. 

> @stanleypliu the images are already in the site somewhere - you can see the ones which are being used on the production site. I'm sure how these are hooked up to the front end at the moment - you may need to have a look into this.

I thought so but couldn't trace where exactly. I'll dig around in the PA fragment for any reference, but all I could find was the attachment blob.

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,103 @@
...
+        position: relative;
+        user-select: none;
+        width: 100%;
+    }
+
+    &__selected {
+        @include flex;
+        @include flex-v-center;
+        @include flex-h-between;
+        @include responsive(padding, rem-calc(14 10 15 18), rem-calc(22 17 29 24), rem-calc(22 17 29 24));
```
stacytalbot, line 53: Height of the select is currently a bit on the large side - on Desktop in the designs it is 69px. Can you also make sure that the text is vertically centered. At the moment the difference in padding top and bottom is too large.

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,103 @@
...
+        display: block;
+        position: relative;
+
+        &:hover {
+            background-color: $grey-light;
+        }
+    }
+
+    &__dropdown {
+        @include icon-chevron-black-down;
```
stacytalbot, line 93: This needs nudging in a bit so it isn't so close to the edge.
Screenshot 2020-07-20 at 12 46 55

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,103 @@
+&--equity {
+    @include responsive(display, block, block, flex);
+
+    &__button {
+        @include flex-vs-start;
+    }
+
+    &__content {
+        @include responsive(width, 100%, 100%, 50%);
+        @include responsive(height, rem-calc(271), rem-calc(507), auto);
```
stacytalbot, line 10: Screenshot 2020-07-20 at 12 46 08
Needs a bit of vertical padding on mobile.

app/assets/stylesheets/components/select/_select-equity.scss
stanleypliu: Sorry, mixed up the order of the padding

app/assets/stylesheets/components/select/_select-equity.scss
stanleypliu: Screenshot 2020-07-20 at 17 01 40

It looks fine to me. Is that on Browserstack?

stacytalbot: No this was in chrome

stacytalbot: Also we are taking out this responsive mixin - sorry I should have mentioned this before. Can you just use the normal mixins `@include breakpoint($small) { }`

app/assets/stylesheets/components/select/_select-equity.scss
stacytalbot: Yea that is the order it is mean to be in.

Can you amend the top and bottom padding - it still isn't vertically aligned. For desktop you need top: 19px and bottom 18px; so that it is also 69px tall with the text vertically centered.

app/assets/stylesheets/components/select/_select-equity.scss
stacytalbot: padding issue is in the mobile breakpoint roughly from 500px upwards

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,174 @@
...
+            margin: 0 auto;                 
+        }
+    }
+
+    &__description {
+        @include flex-grow;
+    }
+
+    &__header {
+        @include breakpoint-down($small) {
```
stacytalbot, line 54: Here you don't need to use the breakpoint-down mixin, you would just set the margin as normal and the it will be overwritten but the margin-bottom set in line 58

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,174 @@
...
+        }
+        @include breakpoint($large) {
+            margin-bottom: rem-calc(14);
+        }
+
+        margin-top: 0;
+    }
+
+    &__panel {
+        @include breakpoint-down($small) {
```
stacytalbot, line 68: Same again here

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,174 @@
...
+        background-color: $grey-xlight;
+    }
+
+
+}
+
+// Custom select for this component
+&--dropdown {
+    &__custom-select {
+        @include breakpoint-down($small) {
```
stacytalbot, line 91: and here

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,174 @@
...
+            margin: rem-calc(14 0 19);
+        }
+
+        position: relative;
+        user-select: none;
+        width: 100%;
+    }
+
+    &__selected {
+        @include breakpoint-down($small) {
```
stacytalbot, line 107: and here

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,174 @@
...
+            position: absolute;
+            top: 100%;
+            left: 0;
+            right: 0;
+            z-index: 2;
+        }
+    }
+
+    &__option {
+        @include breakpoint-down($small) {
```
stacytalbot, line 145: and here

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -0,0 +1,174 @@
+&--equity {
+    @include breakpoint-down($small) {
+        max-width: rem-calc(685);
+    }
+    @include breakpoint-down($large) {
```
stacytalbot, line 5: You dont need to set this within the breakpoint down mixin

app/assets/stylesheets/components/select/_select-equity.scss
```diff
@@ -16,9 +15,6 @@
 
     &__content {
         @include breakpoint-down($small) {
```
stanleypliu, line 17: Thought it was best to leave this in as the image was getting too small with just `object-fit: contain` at larger px.

PR #434, opened 2020-07-17
custom filter checkboxes cms tag created

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/130

Created a custom checkbox component that is designed to be used to populate filters in the News and Stories section. This can be added to the CMS and passed any number of string arguments, as well as whatever name the user wants to add - these arguments are initialised as part of the FilterCheckbox fragment attributes so could probably be accessed using `cms_fragment_content` by the front end or something similiar. Might need a bit of styling, but just want to check that it works as it should first. 

P.S. New branch has been created in the db submodule as I've edited `db/cms_seeds/protected-planet/layouts/template-basic/content.html` to include two of these tags.

### DIFF
--- a/app/assets/javascripts/comfy/admin/cms/filters.js.coffee
+++ b/app/assets/javascripts/comfy/admin/cms/filters.js.coffee
@@ -1,12 +1,18 @@
 $(document).ready( ->
+  # Made it able to support any number of options
+  customCheckboxes = document.getElementsByClassName('filter-checkbox');
+  customElems = Array.from(customCheckboxes).map((elem) => elem.innerText);
+
+
   # Create a change event listener for each topic checkbox.
   # This will add or remove the related substring in the content input tag.
-  ['WDPA', 'OECM', 'PAME'].forEach((el) =>
-    $('#fragment-topic_'+el).change((event) =>
+
+  customElems.forEach((el) =>
+    $('#fragment-topic_' + el).change((event) =>
       $content = $('#fragment-topic_content')
       _val = $content.val()
       if (event.target.checked)
-        $content.val(_val+' '+el)
+        $content.val(_val + ' ' + el)
       else
         $content.val(_val.replace(el, '').trim())
     )
--- a/app/assets/javascripts/comfy/admin/cms/filters.js.coffee
+++ b/app/assets/javascripts/comfy/admin/cms/filters.js.coffee
@@ -4,11 +4,11 @@ $(document).ready( ->
   ['WDPA', 'OECM', 'PAME'].forEach((el) =>
     $('#fragment-topic_'+el).change((event) =>
       $content = $('#fragment-topic_content')
-      _val = $content.val()  
+      _val = $content.val()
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/comfy/admin/cms/filters.js.coffee
```diff
@@ -0,0 +1,14 @@
+$(document).ready( ->
+  # Create a change event listener for each topic checkbox.
+  # This will add or remove the related substring in the content input tag.
+  ['WDPA', 'OECM', 'PAME'].forEach((el) =>
```
stanleypliu, line 4: You're using a hard-coded array here of categories, is it possible that we could pass an array to this function instead dynamically?

stanleypliu: But good call with the JavaScript to change the form.

app/assets/javascripts/comfy/admin/cms/filters.js.coffee
```diff
@@ -0,0 +1,14 @@
...
+    $('#fragment-topic_'+el).change((event) =>
```
stanleypliu, line 5: Where does `#fragment-topic` come from? Doesn't seem to be defined in `filter_checkboxes`, so I assume it's somehow being incorporated into the ID of the element.

lib/cms_tags/filter_checkboxes.rb
```diff
@@ -1,57 +1,39 @@
...
-  end
-
-  def store_in_content(category)
-    value = self.send(category)
-    fragment.content << " #{category} #{value} "
+    @custom_categories = options['options'].split(',').map(&:strip)
   end
   
   def form_field(object_name, view, index)
     name = "#{object_name}[fragments_attributes][#{index}][content]"
```
stanleypliu, line 8: Here's what I don't really understand about the variables inside the square brackets - they're not indices of any array, the square brackets seem to mean something else in this context. Although when I changed `boolean` to `content` in the last bracket, I could tell that it's basically specifying which attribute of the current tag to use.

app/assets/javascripts/comfy/admin/cms/filters.js.coffee
stanleypliu: I've changed this now to be dynamic.

lib/cms_tags/filter_checkboxes.rb
Levia: `fragments_attributes` it's not about array indices, it's basically related to the nested object attributes.
For example, here we have a form for a Page object, to assign values to it's nested object, in this case a Fragment, you can use the above syntax

stanleypliu:
Still need to do a couple of things - try to add CSS to the tag dynamically using JavaScript/Coffeescript as it's misaligned at the moment with the label, and also account for the case when there's more than one of these filter checkbox tags present on the page, so there needs to be a way to identify each tag.

PR #440, opened 2020-07-21
Ticket 33

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/33

Make sure you have redis and sidekiq running.
You will need a MAILER_HOST env variable.
The PDFs do not have working maps yet - this shouldn't stop the merge of the PR.


stanleypliu:
Just added a new helper for the IUCN categories/governance charts on the country/region pages that will populate both sets of links.

stanleypliu:
Download modal wasn't appearing when I clicked on the Download button - turns out the merge conflict was resolved incorrectly so I've pushed the old method.

stanleypliu:
@stacytalbot I've left comments on this PR via https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/33 - front end looks good but it seems like there's problems uploading files to the S3 bucket at times which is required for them to be downloaded. Maybe this is related to the less optimised connection between the development environment and the bucket, and it might not happen in production.

app/assets/stylesheets/components/cards/card/stats/_card-stats-affiliations.scss
```diff
@@ -1,8 +1,19 @@
+//--------------------------------------------------
+// mixins
+//--------------------------------------------------
+@mixin card-stats-affiliations-for-pdf {
+  .pdf & {
```
stanleypliu, line 5: Does this mixin need to be deleted, looks like it's not used?

stacytalbot: Yea i'll delete these

stacytalbot: done

app/helpers/downloads_helper.rb
```diff
@@ -0,0 +1,78 @@
+module DownloadsHelper
+  def download_options options_array, domain, token
```
stanleypliu, line 2: I feel like this could be refactored into a  `case when` statement, or at least it could be condensed into a smaller piece of code especially with the outcome being that a hash is being pushed to the `download_options` array

stacytalbot: @Levia are you ok to sort this

Levia: @stacytalbot That's done now

stanleypliu:
There's still an issue with caching, the OECM download of the shapefile which didn't complete properly last time when I tested appears to be immediately ready when I try to download it again and it takes me to a 404 page for the S3 bucket.

stanleypliu:
Also PDF downloads have another issue, as well as not working for individual sites, when I downloaded the regional PDF for NA, I got the page for Namibia instead, and when I tried to download the polar PDF, it gave me the 404 page instead, even though I could view the polar region page without problems.

PR #441, opened 2020-07-21
made links on the home page route correctly to regions, areas etc.

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/12

Links have been added to the circular cards on the homepage that correctly route to Regions, Countries and Areas depending on the card being clicked.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -80,8 +80,10 @@ class HomeController < ApplicationController
   def levels
     _levels = home_yml[:pas][:levels]
     _levels.map do |level|
-      geo_type = level.delete(:geo_type)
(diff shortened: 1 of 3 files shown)


Levia:
@stanleypliu Had to spend some time understanding what was going on with that yml file, but it seems to be related to the `delete` line for :geo_type being called within the levels block. I ended up pushing the fix myself. So please feel free to merge this if you hare happy.

Just to clarify, using the :icon value was not working well as the value for individual areas had to be `sites`, not `areas`

stanleypliu:
Yeah I couldn't understand what was going on with `_levels` as I couldn't work out how it was able to have a URL yet that key wasn't present in the YML.

stanleypliu:
Yep, works fine.

PR #442, opened 2020-07-23
Social share

defaye:
image

Set the defaults for the Opengraph metatags in the application_controller.
Modify app/controllers/comfy/cms/content_controller.rb if you want to handle more Comfy opengraph-related metadata.
Overrides on a per-controller basis can be done within the controllers themselves - see the marine_controller.rb for an example.

The ComfyOpengraph class is meant to be used to parse Comfy CMS pages with an OpengraphBuilder instance.


lib/modules/opengraph_builder.rb
```diff
@@ -0,0 +1,60 @@
...
+class OpengraphBuilder
+  attr_accessor :data
+
+  # Create an instance e.g. og = OpengraphBuilder.new
+
+  # The initializer accepts a hash with the starting structure for example...
+  #
+  # og = OpengraphBuilder.new(
+  #   { og: { title: 'my title' }, twitter: { card: 'my_card' } }
+  # )
```
lucacug, line 15: I would remove this comment as it is already clear from the code how to initialize this class

lib/modules/opengraph_builder.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  private
+
+  def meta_tags(data, prefix)
+    arr = []
+    data.each do |key, value|
+      arr << format(
+        '<meta property="%<prefix>s:%<key>s" content="%<value>s">',
+        { prefix: prefix, key: key, value: CGI.escapeHTML(value.to_s) }
```
lucacug, line 55: I noticed that the `CGI.escapeHTML` is causing this
Screenshot from 2020-07-24 10-53-24
Not sure you really need that method, maybe just `value.to_s` is enough, but I may be missing something here

defaye: The issue was that if there are no results from locales, it could give you a string that would make the HTML very unhappy. I wonder if there is a simpler escaping method I could use in this case?

defaye: On second inspection, it does seem to be sufficient actually, with just `.to_s` with a broken locale:

`<meta property="og:title" content="translation missing: en.opengraph.defaults.titleadsa">`

lib/modules/opengraph_builder.rb
defaye: I removed the first line of this comment but think it's good to keep line 11-15 just so that it's obvious that the hash should be two-levels in depth beginning with the prefix (e.g. og or twitter)

app/views/partials/social/_sharing.html.erb
```diff
@@ -1,3 +1,15 @@
-<%= create_sharing_twitter_link %>
-<%= create_sharing_facebook_link %>
-<%= create_sharing_linkedin_link %>
\ No newline at end of file
+<%= link_to "", "https://twitter.com/intent/tweet/?url=" + request.original_url, 
+        title: url_encode("Share on Twitter"),
```
stacytalbot, line 2: We are moving over to make sure all text is either in the CMS/DB or a yml file so i think these titles can go in a yml file.

config/locales/thematic_area/marine/en.yml
```diff
@@ -15,3 +15,23 @@ en:
         chart_text: "%{country} and its %{overseas} overseas countries and territories have a combined area of %{km}km<sup>2</sup>"
         chart_stat_text: of their national waters are protected
         chart_stat_text_overseas: of their overseas territories waters are protected
+
+      social:
+        title: Explore the World's Marine Protected Areas
```
stacytalbot, line 20: The title and the description were bits I wanted pulling from the CMS so that they can update them and also they will need to be added for all the news pages etc.

Screenshot 2020-07-27 at 15 16 33

lib/modules/comfy_opengraph.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
```
stacytalbot, line 1: haha I can't check this bit - but my question here is does it check to see if the page has a @cms_page object and then check for social-title and social-description fields

app/views/partials/social/_sharing.html.erb
```diff
@@ -1,3 +1,15 @@
...
-<%= create_sharing_facebook_link %>
-<%= create_sharing_linkedin_link %>
\ No newline at end of file
+<%= link_to "", "https://twitter.com/intent/tweet/?url=" + request.original_url, 
+        title: url_encode("Share on Twitter"), 
+        class: "social__icon--twitter", 
+        target: "_blank"
+%>
+<%= link_to "", "https://facebook.com/sharer/sharer.php?u=" + request.original_url,
+        title: url_encode("Share on Facebook"),
```
stacytalbot, line 7: They also don't need to be encoded.

stacytalbot: Screenshot 2020-07-27 at 15 23 14

config/locales/thematic_area/marine/en.yml
defaye: It should be doing this, it depends on whether they have been set or not~

The bits loaded from locales are default settings

lib/modules/comfy_opengraph.rb
defaye: Have a look at `app/controllers/comfy/cms/content_controller.rb` to see where this class gets utilised.
The controller is the overridden controller from Comfy - I've made it obvious where I've made modifications with the comment blocks. Where a page is loaded, it will always have `@cms_page` set, if this helps clarify things?

config/locales/thematic_area/marine/en.yml
stacytalbot: I have just tested this on the Marine page but I get the default ones instead. 

Screenshot 2020-07-27 at 16 19 09
Screenshot 2020-07-27 at 16 18 53

defaye: Ah, this is because marine_controller.rb has a bespoke override, which stems from what was there before my work on this branch began, so it would perhaps be redundant to have this custom override now: https://github.com/unepwcmc/ProtectedPlanet/blob/ticket-54/app/controllers/marine_controller.rb#L27

This can probably be removed if the content is now in the CMS directly.

app/controllers/comfy/cms/content_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def load_seeds
+    return unless ComfortableMexicanSofa.config.enable_seeds
+    ComfortableMexicanSofa::Seeds::Importer.new(@cms_site.identifier).import!
+  end
+
+  # Attempting to populate @cms_page and @cms_layout instance variables so they
+  # can be used in view helpers/partials
+  def load_cms_page
```
stacytalbot, line 70: There is also a method here about populating @cms_page https://github.com/unepwcmc/ProtectedPlanet/blob/555a316947d62e1483efcd5488bc449d5cd4a77a/app/controllers/application_controller.rb#L107

config/locales/thematic_area/marine/en.yml
defaye: If this is redundant, I can remove it and the locale references too? Should theme_image be set for the twitter:card as well? I am sure that Twitter will use og:image by default if not provided though.

config/locales/seo/defaults/en.yml
```diff
@@ -0,0 +1,9 @@
+en:
+  seo:
+    defaults:
+      site_title: Protected Planet
```
stacytalbot, line 4: Could we merge this into the file above? It would be better if we just had 1 set of default site titles and descriptions etc.

app/controllers/comfy/cms/content_controller.rb
defaye: Interesting... it only appears to be used in the marine controller. Do you think it's redundant now?

defaye: Nevermind, I see it being called in the app controller as well. I need to check what it's doing

config/locales/thematic_area/marine/en.yml
stacytalbot: Yea you can take those marine ones out.

Yea i suppose there is no need to set the image twice for twitter.

app/controllers/comfy/cms/content_controller.rb
stacytalbot: Yea so for all the pages that have a controller but also a page in the CMS it is called so that we can populate the page. So likely to be used across a lot of pages.

defaye: If all it does is expose `@cms_page`, I can perhaps look into doing that from the app/controllers/comfy/cms/content_controller.rb instead?

config/locales/seo/defaults/en.yml
defaye: Sure

PR #443, opened 2020-07-23
add search pa zoom

WKocur:
Zooms and adds popup if PA (not working for region or country yet as the search doesn't)

### DIFF
--- a/app/javascript/components/map/mixins/mixin-bounding-box.js
+++ b/app/javascript/components/map/mixins/mixin-bounding-box.js
@@ -7,7 +7,15 @@ export default {
     }
   },
 
+  created () {
+    this.addZoomEventListener()
+  },
+
   methods: {
+    addZoomEventListener () {
+      this.$eventHub.$on('map:zoom-to', this.zoomTo)
+    },
+
     initBoundingBoxAndMap () {  
       if (this.mapOptions.boundsUrl) {
         axiosGetWithoutCSRF(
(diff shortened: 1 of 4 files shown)


app/javascript/components/map/mixins/mixin-bounding-box.js
```diff
@@ -31,23 +39,41 @@ export default {
...
 
-    getExtentResponseZoomToHandler (padding) {
+    getZoomAndPopupHandler (options) {
       return res => {
         const extent = res.data.extent
-  
-        if (extent) {
-          this.map.fitBounds(this.getBoundsFromExtent(extent, padding))
+
+        if (!extent) { return }
```
defaye, line 53: Would ESLint complain if it were `if (!extent) return` instead?

app/javascript/components/map/mixins/mixin-pa-popup.js
```diff
@@ -27,16 +27,19 @@ export default {
...
 
         return pa !== null
       }
     },
 
     addPopup (coords, pa) {
-      const html = pa.wdpaid ? 
-        `<a href="/${pa.wdpaid}">${pa.name}</a>` :
+      const html = pa.url ? 
+        `<a href="${pa.url}">${pa.name}</a>` :
```
defaye, line 42: I'm guessing this will need more work later on

app/javascript/components/map/mixins/mixin-bounding-box.js
WKocur: Probably not, but it is what we do in general and I think my preference, but only slightly so. I don't mind which particularly, so happy to change if you have stronger feelings than I do!

defaye: I'm not too fussed either, I used to really not enjoy seeing inline ifs at all. 

I prefer the return to be dropped to its own line if you want to have braces, I like seeing both inline (but without braces) or not inline but on a new line

PR #445, opened 2020-07-24
Sync seeds

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/136

Wrote a rake task that will copy files from the staging instance via SCP by checking the timestamp of each file and then only downloading it if it's newer. This is flawed in that the files on the server, while technically being newer, might not actually be the latest version as it is fully dependent on the branch that is currently deployed to staging...so we might want to think about checking the versions of the files instead of just relying on modification timestamps. Bonus though, for the CMS files, they will download any to your local machine which don't already exist, such as PDF files.

`rake comfy:staging_import`


lib/tasks/staging_seeds.rake
```diff
@@ -0,0 +1,49 @@
...
+
+namespace :comfy do
+    desc "Import CMS Seed data from staging"
+
+    task :seed_import => :environment do |_t|
+      require 'net/ssh'
+      require 'net/scp'
+      
+      def check_timestamp(file, session, local_file, remote_file)     
+        newer = Time.at(file.attributes.mtime) >= File.stat(local_file).mtime
```
Levia, line 11: I would rename this variable `is_newer` so that's immediately understandable it's a boolean flag

lib/tasks/staging_seeds.rake
```diff
@@ -0,0 +1,49 @@
...
+        if newer 
+          puts "#{file.name} is newer than the local copy, downloading..."
+          session.scp.download!(remote_file, local_file)
+        end
+      end
+      
+      # Locally stored seeds - assumes you already have the local folder - 
+      # it won't create it 
+      local = ComfortableMexicanSofa.config.seeds_path + '/protected-planet'
+      remote = 'ProtectedPlanet/current/db/cms_seeds/protected-planet'
```
Levia, line 21: These two could be constants, also to make sure they won't change by using `.freeze` at the end. I would suggest to define those outside the `task` block at the top, if that works

lib/tasks/staging_seeds.rake
```diff
@@ -0,0 +1,49 @@
...
+      
+      puts "Importing CMS Seed data from Staging Folder to #{local} ..."
+
+      # SSH into staging server with Net::SSH
+      Net::SSH.start(ENV['PP_STAGING'], ENV['PP_USER']) do |session|
+        session.sftp.dir.glob(remote, '**/*').each do |file|
```
Levia, line 27: I wonder if we should consider pulling entire folders based on timestamps instead of files?
I mainly see 2 reasons for that:
 1. Reduce the number of scp commands
 2. Files can be deleted from staging so I don't think the script is taking this into account. This means that the system that is pulling data from staging might potentially store old and unnecessary files forever.

lib/tasks/staging_seeds.rake
```diff
@@ -0,0 +1,49 @@
...
+            puts "#{file.name} doesn\'t exist locally, downloading"
+            # File doesn't exist locally, so download it (in any folder required)
+            session.scp.download!(remote_file, local_file)
+          end
+        end
+      end
+
+      puts "Finished downloads, now replacing your local seed data..."
+
+      Rake::Task["comfy:cms_seeds:import"].invoke('protected-planet', 'protectedplanet')
```
Levia, line 47: Just a reminder to double check this works properly when migrating to AWS with ActiveStorage

lib/tasks/staging_seeds.rake
stanleypliu: True, I could just download the folders recursively instead of singly, but not sure I'm able to pass a block to the `download!` method as it needs to check the timestamps and also make sure the encoding is correct. And yes, the script isn't taking deleted files into account, I should really add that functionality in.

Levia: I think that `download!` method might simply accept folder names as it does for files? In the same way you should be able to check timestamps for the folders. I believe that if something within a folder changes (e.g. another folder or a file), the timestamp of that folder changes too

stanleypliu: It does indeed, OK I'll just move all of that logic into the function and see how it goes.

stanleypliu: Task seems to work as before, but I've noticed something - we have two copies of every file in the `files` folder, one with a preceding underscore and one without. Is this something we want? I thought this was an error.

Screenshot 2020-07-28 at 18 17 33

stanleypliu:
@Levia Give this another try when you have the time - I've tried my best today to clean this up and separate out the code blocks into functions, but it is still a bit of a mess 🤕. However, there is a crude router in place to allow users to select what to download (everything, or just select folders).

Gemfile
```diff
@@ -30,7 +30,9 @@ gem 'vuejs-rails', '~> 2.3.2'
 gem 'sprockets-vue', '~> 0.1.0'
 
 gem 'rails-controller-testing'
-gem 'gdal', '~> 2.0'
+# gem 'gdal', '~> 2.0'
+# TODO Stanley - Need to change this version of gdal back to 2.0 (and remove from gitignore) when I reinstall GDAL to 2.4.4
+gem 'gdal', '~> 3.0'
```
Levia, line 35: @stanleypliu Could you please make sure to restore this as well?

stanleypliu: Sorry, I'll restore it on every branch!

lib/tasks/staging_seeds.rake
```diff
@@ -0,0 +1,63 @@
...
+  PP_USER = 'wcmc'.freeze
+
+  desc "Import CMS Seed data from staging"
+  task :staging_import => :environment do |_t| 
+    
+    puts "Importing CMS Seed data from Staging Folder to #{LOCAL} ..."
+
+    puts question = "What would you like to import? 'All/Files/Layouts/Pages' or 'Nothing' to quit"
+    valid_answers = ['All', 'Files', 'Layouts', 'Pages', 'Nothing']
+    answer = STDIN.gets.chomp.downcase.capitalize
```
Levia, line 20: I think it would be better to have one of those passed as an argument (downcased) to the task instead of asking the question at runtime as it would be quicker and checks would be easier. What do you think? @stanleypliu

lib/modules/sync_seeds.rb
```diff
@@ -0,0 +1,126 @@
...
+    puts "Downloading #{source} as it's newer or not found locally"
+    return @session.scp.download!(source, dest) if File.file?(dest)
+    @session.scp.download!(source, dest, recursive: true)
+  end
+  
+  def create_paths(relative_path)
+    { local_path: File.join(LOCAL, relative_path), remote_path: File.join(REMOTE, relative_path) }
+  end
+
+  def check_if_newer(parent_folder, local_item, remote_item, remote_path, local_path, base = LOCAL)
```
Levia, line 52: I notcied the import taks always downloads the same files regardless if I just run it once before.
After debugging this a bit I noticed this goes to the `else` statement of the first condition, so saying that the resource doesn't exist. I wonder if it's because of the first parameter passed to this method on line 104 perhaps? @stanleypliu

lib/tasks/staging_seeds.rake
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+namespace :comfy do
+  
+  # Locally stored seeds - assumes you already have the local folder - 
+  # it won't create it 
+  LOCAL = (ComfortableMexicanSofa.config.seeds_path + '/protected-planet').freeze
+  REMOTE = 'ProtectedPlanet/current/db/cms_seeds/protected-planet'.freeze
+  PP_STAGING = 'new-web.pp-staging.linode.protectedplanet.net'.freeze
+  PP_PRODUCTION = 'new-web.pp-production.linode.protectedplanet.net'.freeze
```
Levia, line 10: We are not really using this at the moment as we are mainly testing on the local environment.
Perhaps you could make a condition checking the `Rails.env` and using the local address or PP_PRODUCTION based on that?

stanleypliu: Hmm, how about I could use a dynamic variable for the remote folder like so: 
`'new-web.pp-#{Rails.env}.linode.protectedplanet.net'`

stanleypliu:
Commented out as I don't want to modify my own database with staging given that it's not always the most up to date. About running the import at the same time as the download, do you mean it should work like the following example: say someone just chooses to download layouts, then I could also invoke  the associated Comfy task like so? `Rake::Task["comfy:cms_seeds:import"].invoke('protected-planet', 'protectedplanet', 'layouts')`

lib/tasks/staging_seeds.rake
stanleypliu: I personally have always liked the idea of a router interface as it feels more natural to me to input a command and then have the program execute what I just typed in. Plus, it also becomes more apparent to the user what can be passed to the task and what can't, as the task will inform them if their input is invalid.

lib/tasks/staging_seeds.rake
Levia: I'm not sure I fully understand here.
So this taks should be used also so that production can get the seeds from staging and import those when ready.
So both when running this task in development and production shoul use the staging server as the main reference.

lib/tasks/staging_seeds.rake
Levia: Ok, we can leave it like that. I was a bit concerned about the fact that people could run this task assuming it already does everything and leave, without noticing an input is required :) But it's fine like this as well

lib/tasks/staging_seeds.rake
stanleypliu: Ah I see, I was always under the impression that we'd want to obtain the content from production because that's the one that gets updated most frequently and has the right content by CLS...or is it not the case with Protected Planet?

Levia: The main reason for this work I'm aware of, is that the CMS will get first populated on staging so not to disrupt production and so that we could do all the necessary checks there before deploying live. This mean that once everything is ready on staging with regards the CMS, we would need to run this task on production to fetch the CMS seeds from staging and also run the import task. Hope it makes sense

Levia:
@stanleypliu Yeah it's not really mandatory, but perhaps a nice to have.
So I'm imagining a rake task that, when invoked, it would run this one you just made here, asking also for what would you like to import, and then, when that bit is done, it would also invoke the import example automatically.

lib/tasks/staging_seeds.rake
stanleypliu: You might have a point actually, if we want to run this on production, do we want to automate this so that it'll function without having to wait for user input?

Levia: Potentially yeah.
You could even make it work both ways; for example you could make a check where, if arguments are provided, you won't ask for user input, while it does ask otherwise.

lib/modules/sync_seeds.rb
stanleypliu: It's due to the encoding, I thought I'd managed to squash all the issues to do with filenames that have non-ASCII characters but evidently some still slip through the cracks. This is only affecting the files folder for obvious reasons. I'm going to fix this.

lib/tasks/staging_seeds.rake
stanleypliu: That's the approach I went for ^

lib/modules/sync_seeds.rb
```diff
@@ -0,0 +1,137 @@
...
+      end
+    end
+  end
+
+  def commence_import(answer)
+    logger = ComfortableMexicanSofa.logger
+    ComfortableMexicanSofa.logger = Logger.new(STDOUT)
+
+    if answer == 'All'
+      Rake::Task["comfy:cms_seeds:import"].invoke('protected-planet', 'protectedplanet')
```
Levia, line 128: I feel like it might be useful to have the folder name and site name be passed as arguments to the task. This is so you can change this when required at the moment of running the task without changing the code. This would also affect the `LOCAL` constant in the rake task 

Worth also adding a description/comment above the task on how to run if after the changes (like you just did for the seed_type)

lib/modules/sync_seeds.rb
```diff
@@ -0,0 +1,137 @@
...
+
+  def commence_import(answer)
+    logger = ComfortableMexicanSofa.logger
+    ComfortableMexicanSofa.logger = Logger.new(STDOUT)
+
+    if answer == 'All'
+      Rake::Task["comfy:cms_seeds:import"].invoke('protected-planet', 'protectedplanet')     
+    else
+      module_name = "ComfortableMexicanSofa::Seeds::#{answer.singularize}::Importer".constantize
+      module_name.new('protected-planet', 'protectedplanet').import!
```
Levia, line 131: See previous comment

lib/tasks/staging_seeds.rake
```diff
@@ -0,0 +1,62 @@
...
+        end
+
+        puts "Downloading a new set of #{answer.downcase}..."
+
+        new_session.main_task(local_list, remote_list)
+      end
+
+      puts "Finished downloads, now replacing your local seed data with your selection..."
+
+      new_session.commence_import(answer)
```
Levia, line 58: I was under the impression this would have to go into a different task altogether with staging_import one?
The idea was to leave this task to just fetch the data from the server and then create another rake task that would do both things. The alternative would be to pass another argument to this task but I feel it might be too much.

Also I think it might be a good idea to first check that new files have been actually fetched from the server before running the importer, so it doesn't run unnecessarily if everything is already up to date. Even though it seems the importer is smart enough to notice changes by itself while running. What do you think?

lib/modules/sync_seeds.rb
stanleypliu: So my rake task would be invoked like the Comfy task with 'protected-planet', 'protectedplanet'?

Levia: Potentially yeah. Or you could check whether the folder name would be enough on it's own, meaning that at the moment of importing the folder name should be exactly the same as the site name on the CMS

lib/tasks/staging_seeds.rake
stanleypliu: Re the different task, when I first conceived how I would go about this Rake task I was thinking to just download them like you said, and then give the user the option to run the import if they wanted to. I guess it came down to my assumption that if you're intending to run this task you're doing so with the intention of wanting to update your database as well, so I thought I'd just save time by importing automatically.

Re the second point, that is a good shout. However, in an earlier comment, you addressed the point of whether I should download individual files if they're missing or older, or the entire folder to reduce the number of `scp` commands being run. Is that still valid? To make the importer smarter it could just download files instead of the whole folder again in the case of just one file or folder (in `files`, `pages` etc - just has to be in one of the directories in `cms_seeds`) - at the moment I've put in a `break` to stop it from downloading the whole folder again if `check` returns `true`, but I feel that it is a more natural way to just check each file individually.

Levia: Yeah, I guess we could leave it as it is then. Not particuarly relevant as you say.

I'm not sure I follow abou the second point. What I had in mind was about adding a flag to store the result of the whole fetching process from the staging server, if possible. If there have been changes then launch the import as well, don't otherwise. But again, this is also a minor thing considering that the importer seems to deal with this already in a way

stanleypliu:
@stacytalbot if you want to test the version of my script as it is at the moment, on your local branch you can copy the `cms_seeds` folder in your local branch (backing up your old copy) so that the script will only modify the dummy folder you've just created so it won't erase any changes you've made.

stanleypliu:
@Levia Can we merge this in now, and if there is any change needed, we can add it later? I think the intention is for us to use it whilst we're working on the project to get the latest changes.

Levia:
There might be some minor amendments to do still but could be done at a later stage. Merging it now!

PR #449, opened 2020-07-28
Add tabs concern refactor

defaye:
More of a learning exercise for me to understand concerns, but could see this little refactor makes sense?


app/controllers/concerns/tabs.rb
```diff
@@ -0,0 +1,18 @@
+module Concerns::Tabs
+  extend ActiveSupport::Concern
```
stanleypliu, line 2: I think this isn't needed as you're not relying on any other module as a dependency, so it can be a standard module - @Levia what do you think?

stanleypliu:
@defaye I'm going to make this a helper if that's alright - the reason being is that this is just a single method that's used in controllers rather than models, hence you wouldn't be able to take advantage of class methods, which is something that a concern offers, nor do you need any other module for your method to work, which is what the `included` does that's part of a concern. It doesn't mean that this doesn't work, it's just that in keeping with the rest of the repo

stanleypliu:
Actually, I did retain the concern in the end, just refactored to include the CmsHelper module to change how the content from the fragment was fetched.

PR #450, opened 2020-07-28
Location search

WKocur:
This adds extent_url method to country and region models.

Also only adds oecm filter to search if 'oecm' or 'wdpa' is the db_type. Otherwise filters = {}

You can test this on the map search on the homepage.

BUG: I can't get OECMs to be found either if I set db_type = oecm or db_type = site (therefore filters = {})...


config/locales/main_map/en.yml
```diff
@@ -22,8 +22,8 @@ en:
         title: Region
         placeholder: Search...
       -
-        id: wdpa
-        title: Protected Area
+        id: site
```
stacytalbot, line 25: Can you pull the config out of the yml file?

PR #451, opened 2020-07-28
Add new component for related countries

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/131


Levia:
@stacytalbot conflicts here as well, please

app/views/partials/stats/_stats-related-countries.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <h2><%= t('stats.related-countries.title') %></h2>
+
+  <% if country_children %>
+    <h3 class="stats__subtitle"><%= t('stats.related-countries.title-child') %></h3>
+    <ul class="list--underline">
+      <% country_children.each do |child| %>
+        <li class="list__li">
+          <span class="list__title"><%= child['name'] %></span>
+          <span class="list__value"><%= child['iso_3'] %></span>
+          <%= link_to 'View', country_path(child['iso']), title: '', class: 'list__right list__a' %>
```
Levia, line 11: Shouldn't this (and also below) be `child['iso_3']` instead of `child['iso']`? @stacytalbot

Levia: It works either way really. I noticed we are a bit inconsistent when we redirect to the country pages.
From the PA page we use `iso` everywhere (so the url will be /country/IT) while from the areas search we use `iso_3` (/country/ITA). I think iso_3 would be better everywhere, what do you think? @stacytalbot

stacytalbot: yea i can update - i was just reworking what was here originally.

PR #452, opened 2020-07-30
Ticket 69

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/69

Finished black overview boxes on Region, Country and Site pages.


Levia:
@stacytalbot Could you please fix the conflicts when you get the change?

Levia:
@stacytalbot I took the chance here to populate the number of OECM areas for countries and can also do for regions.
However, I wonder if the first number about "Total Protected Areas" should be about WDPAs + OECMs or just WDPAs ?

PR #453, opened 2020-07-30
Cms custom categories

Levia:
## Description

⚠️ This is still a WIP
This is about allowing the creation of custom categories to attach to layouts.
Each page inheriting a layout with the custom categories can be categorised with the ticked categories.

This also includes the ability to use ElasticSearch to filter the pages by those custom categories.
An important note: the categories groups and values will be generated in the backend so are not fully customisable by people using the CMS; this means developers input is required to add more categories groups and values. One of the reason for this is that we are allowing categories to be localised and translated in other languages.

## Notes

It needs the related db branch : https://github.com/unepwcmc/protectedplanet-db/pull/37


app/helpers/cms_helper.rb
```diff
@@ -88,6 +88,40 @@ def get_filtered_pages pages
...
+        {
+          id: pc.id,
+          name: localised_pcs[pc.label.to_sym]
+        }
+      end
+
+      # frontend should return the list of selected categories as follows:
+      # 'group_name' => [category_ids] ; e.g. 'topics' => [1,2,3]
+      {
+        name: name,
```
stacytalbot, line 119: @Levia this needs to be translatable - i believe this is 'Topics' or 'Types',

Levia: @stacytalbot That's done now, I've changed the structure of the yml slightly to include the translatable name

PR #455, opened 2020-07-30
Migrating ActiveStorage files to S3

stanleypliu:
Currently, staging and production servers use 'local' ActiveStorage (as in there is a folder called 'storage' that contains all of the ActiveStorage files) to oversee all the files that are uploaded via the CMS and used on each page (like the various photos, PDFs etc.). They could be backed up to S3 to allow syncing between staging and production to happen more easily, and also sync up with our local repos as well (as we each have a 'storage' folder ourselves which contains all of the ActiveStorage files.

This changes the location of the ActiveStorage files locally on staging as well  to `ProtectedPlanet/shared/storage` (see the commented out line in storage.yml for the new location) to avoid it being overwritten by deploys. I've already migrated the existing Comfy files which have a valid ActiveStorage equivalent to the new shared folder as well as the S3 bucket. Now future uploads should not be overwritten either.


lib/modules/s3.rb
```diff
@@ -30,23 +30,17 @@ def self.link_to file_name, opts={for_import: false}
...
 
   def upload object_name, source, opts
-    # Check to see what environment Rails is currently running in
-    bucket = Rails.application.secrets.aws_downloads_bucket
-
-    if Rails.env.staging? || Rails.env.production?
-      bucket = Rails.application.secrets.aws_files_bucket
-    end
+    # Default to downloads bucket unless specified (e.g. for when uploading CMS files)
+    bucket = opts[:bucket] || Rails.application.secrets.aws_downloads_bucket
```
stanleypliu, line 34: It looks you've partly reverted my change. Did it not work with the secrets?

Levia: So this method have been used to upload downloads files on S3 so far, so using the aws_downloads_bucket.
Now it would also need to use another bucket for uploading the CMS files, so yeah it was necessary to make this change in the end.

With your latest code though, it was using the aws_downloads_bucket on staging or production regardless if the upload intent was for CMS files or for downloads; while using the files bucket by default if it was development instead

lib/tasks/copy_to_s3.rake
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    puts "Could not migrate records with IDs: #{unmigrated_files.join(', ')}"
+  end
+
+  desc "Sync up production's ActiveStorage files with staging's"
+  task :sync_staging_production => :environment do |_t|
+
+    # TODO: When ActiveStorage is set up on production, try to get it synced up 
+  end
```
Levia, line 52: @stanleypliu What should we do about this one?
Do you think we could simply copy the cms_seeds folder onto the production server and run the comfy import task from there? It's a bit more manual but perhaps it could save us some development and testing time given this should only be a one off.

stanleypliu: I think we could still make it a Rake task to copy the folder onto production and do everything else, and then delete this file once we're done with everything. I like the convenience of being able to run it from the command line.

Levia: Ok, so let's leave this for later perhaps so we can have this merged and tested on staging in the meantime. What do you think?

stanleypliu: You mean this task or this branch in general? I think we need this branch on staging so that it will store files in a safe location.

Levia: Yeah I meant this currently empty task only. So that we can merge this branch in the meantime and implement that later

stanleypliu: Yeah I'm good for that.

config/storage.yml
```diff
@@ -5,3 +5,20 @@ test:
 local:
  service: Disk
  root: <%= Rails.root.join("storage") %>
+
+# Main service to be used for staging - also backs up to prod
+staging:
+  service: S3
+  access_key_id: <%= ENV["AWS_ACCESS_KEY_ID"] %>
+  secret_access_key: <%= ENV["AWS_SECRET_ACCESS_KEY"] %>
+  region: us-east-1
```
Levia, line 14: @stanleypliu I think this and the region on production config need to change now too, right?

Levia:
Made some quick amendments to the rake task to account also for fragments attachments. Merging now

PR #458, opened 2020-07-31
Ticket 91 + Ticket 88

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/91
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/88

Sorry I ended up combining two tickets by mistake in the same branch. The second ticket concerns the addition of the related search results partial as well as the link to take the user to the search results page with the relevant filters pre-selected. The second ticket concerns transboundary sites, and a transboundary filter has been added to the results page to take the user to all the transboundary sites if the current site is itself transboundary.


Levia:
@stacytalbot @stanleypliu I can't really understand the merge conflicts that arised here. Can any of you double check and fix this accordingly please?

app/controllers/protected_areas_controller.rb
```diff
@@ -90,10 +94,25 @@ def load_other_designations
...
+  # end
+
+  def get_other_sites
+    return ProtectedArea.all_except(@protected_area).take(3) if @countries.length <= 1
+    ProtectedArea.all_except(@protected_area).transboundary_sites.take(3)
+  end
+
+  def determine_search_path(area)
+    if area.is_transboundary
+      search_areas_path(geo_type: 'transboundary')
```
Levia, line 113: @stanleypliu  This is incorrect as the geo_type attribute is about sites VS country VS region, so it throws an error.
At most, this should either be the same as the url in the `else` statement, or should assign a filter similar to the greenlist one like: `filters[special_status][]=is_transboundary`.

@stacytalbot I remember that we decided to not include this filter in the search at some point; does this one need to come back now that we have the functionality?

Levia:
@stanleypliu While we get an answer for the comment I just made, please be aware that I've added some quick commits to this branch, so you would need to pull before making further changes

app/controllers/protected_areas_controller.rb
stanleypliu: Admittedly I was unsure where it would go, my first thought was that it should be a special status, but then I thought that it's basically a variation of a site, so I classified it under `geo_type`

stanleypliu: @stacytalbot @Levia I've added transboundary as a special status now, instead of a `geo_type`

app/controllers/protected_areas_controller.rb
```diff
@@ -92,10 +96,25 @@ def load_other_designations
...
+  # TODO: Methods, models, controllers, modules related to Networks are slated for removal
+
+  # TRANSBOUNDARY_SITES = "Transboundary sites".freeze
+  # def load_networks
+  #   networks = @protected_area.networks.reject(&:designation)
+  #   # ensure that transboundary sites network always appears first
+  #   networks.sort { |a,b| a.name == TRANSBOUNDARY_SITES ? -1 : a.name <=> b.name }
+  # end
+
+  def get_other_sites
```
Levia, line 108: @stanleypliu 2 points here:

1. I would use `if @protected_area.is_transboundary` as it already contains that kind of logic
2. I wonder if we should return other areas within the same country if not transboundary, rather than any other areas around the world? For example, it feels a bit odd that, if I'm looking at a PA in Italy, I see other 3 sites which are in a completely different country

Point 2 is perhaps more of a question for @stacytalbot

stacytalbot: Marine stated - if a site is transboundary - show 3 other sites that are transboundary, they don't need to be in the same country. Otherwise we may end up with this being empty if there is only 1 transboundary site in a country

Levia: Thanks @stacytalbot . My question was more about when looking at a site which is 'not' transboundary.
If 'not' transboundary, the query below gets any 3 sites which are not the one I'm looking at; so those sites could be anywhere.

It works well already if the site I'm looking at 'is' transboundary

stacytalbot: Sorry. Yes for non-transboundary sites you should show 3 more sites in the same country.

stanleypliu: That's been done - should return sites in the same country now if not transboundary.

app/controllers/green_list_controller.rb
```diff
@@ -93,6 +98,10 @@ def find_protected_area
     @protected_area or raise_404
   end
 
+  def get_green_list_sites
+    @example_greenlist ||= ProtectedArea.green_list_areas.take(3)
```
Levia, line 102: @stanleypliu I guess you could use the `green_list_areas` method below instead of rewriting the same `ProtectedArea.green_list_areas` logic?

PR #459, opened 2020-07-31
Add source csv to zip on shp downloads

lucacug:
This is to add the source table as csv to the zip file whne users download shp. Previously the source table was included only when downloading csv format

[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/70)

### DIFF
--- a/lib/modules/download/generators/shapefile.rb
+++ b/lib/modules/download/generators/shapefile.rb
@@ -104,6 +104,11 @@ class Download::Generators::Shapefile < Download::Generators::Base
     range = (0..@number_of_pieces-1)
     files_paths = range.map { |i| zip_path(i) }.join(' ')
     system("zip -j #{zip_path} #{files_paths}") and system("zip -ru #{zip_path} *", chdir: ATTACHMENTS_PATH)
+    system("zip -ru #{zip_path} #{File.basename(sources_path)}", chdir: File.dirname(sources_path))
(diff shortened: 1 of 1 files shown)


lib/modules/download/generators/shapefile.rb
```diff
@@ -104,6 +104,11 @@ def merge_files
     range = (0..@number_of_pieces-1)
     files_paths = range.map { |i| zip_path(i) }.join(' ')
     system("zip -j #{zip_path} #{files_paths}") and system("zip -ru #{zip_path} *", chdir: ATTACHMENTS_PATH)
+    system("zip -ru #{zip_path} #{File.basename(sources_path)}", chdir: File.dirname(sources_path))
     range.each { |i| FileUtils.rm_rf(zip_path(i)) }
   end
+
+  def sources_path
+    File.join(File.dirname(zip_path), "WDPA_sources.csv")
+  end
```
Levia, line 113: It seems like this method already exists in the base class this one inherits from?
Perhaps you could just assign `zip_path` to an instance variable with the same name so that it can be used by the method in the base class? @lucacug

Levia: @lucacug Also just realised it's worth considering using the already existing method in the base class named `export_sources`, which creates the file first if it doesn't exist at the moment of generating the download

Levia:
This has been fixed within the refresh work. Closing

PR #464, opened 2020-08-05
Fact cards

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/120

Screenshot 2020-08-05 at 11 58 42

Need to check whether the calculations are accurate, in particular the % OECM coverage, which I'm told, can either be considered as separate from PAs, or not. For the purposes of the card, I've considered them separately from PAs, so the %s of the last two cards are actually the total land, or total marine area of the PAs marked as OECMs as a percentage of the **global** land area (not PA land area, I used `CountryStatistic.global_<type>_area` to calculate this). If this is not correct, let me know.


Gemfile
```diff
@@ -35,7 +35,8 @@ gem 'vuejs-rails', '~> 2.3.2'
 gem 'sprockets-vue', '~> 0.1.0'
 
 gem 'rails-controller-testing'
-gem 'gdal', '~> 2.0'
+
+gem 'gdal', '>= 2.0'
```
Levia, line 39: @stanleypliu I think this not really required right?

app/models/protected_area.rb
```diff
@@ -147,6 +147,18 @@ def overlap(pa)
     overlap
   end
 
+  def self.global_terrestrial_oecm_coverage
+    land_oecms = ProtectedArea.where(is_oecm: true, marine: false).pluck(:reported_area)
```
Levia, line 151: @stanleypliu Here and also in the method below, you could make use of the scope already implemented in this class instead of the where statement, so you can also concatenate them.
e.g.

```
self.oecms.marine_areas.pluck(...)
```

Levia:
@stacytalbot If you could please confirm that the one described above should be the desired behaviour, I'm generally happy with this ones the minor comments are addressed

Gemfile
stanleypliu: I have to have it for myself 😞 since I'm still searching for a Homebrew GDAL package v2.4

Levia: Ok, so you can keep it for your local environment but this is not actually a requirement for staging/production and other local enviroments so it doesn't need to be pushed on the repo

app/models/protected_area.rb
```diff
@@ -147,6 +147,18 @@ def overlap(pa)
     overlap
   end
 
+  def self.global_terrestrial_oecm_coverage
+    land_oecms = ProtectedArea.oecms.marine_areas.pluck(:reported_area)
```
Levia, line 151: @stanleypliu Here you are now fetching marine areas instead of terrestrial. If there's not terrestrial scope you can create one. Also, since you are already in the `ProtectedArea` class, you can use `self` instead of `ProtectedArea`, as `self` will refer to the class itself given these are class methods

stanleypliu: 😓 that's what I get for not checking after I copy and paste

PR #467, opened 2020-08-05
Dopa link

stanleypliu:
**Ticket**

Associated PR for migration: https://github.com/unepwcmc/protectedplanet-db/pull/39

You'll need to run `Wdpa::DopaImporter.import` in the Rails console to import the CSV, as it's pretty large expect it to take a number of minutes. 

Not sure what's going on with the previous commits - I thought this was a new branch.


lib/modules/wdpa/dopa_importer.rb
```diff
@@ -0,0 +1,32 @@
...
+      ProtectedArea.update_all(is_dopa: false)
+      logger = Logger.new(STDOUT)
+
+      pas_to_update = []
+
+      # Read the CSV to get the WDPA IDs
+      CSV.foreach(DOPA_LIST, headers: true) do |row| 
+        # Find the associated Protected Area with each WDPA ID mentioned and update the is_dopa column to true
+        dopa_pa = ProtectedArea.where('reported_area > 0.5e1')
+                  .or(ProtectedArea.where('reported_marine_area > 0.5e1'))
```
Levia, line 17: Why is this check necessary? I think we would need to find the PAs by the WDPA ID only.

lib/modules/wdpa/dopa_importer.rb
```diff
@@ -0,0 +1,32 @@
...
+
+        if dopa_pa.nil? 
+          logger.info "DOPA site #{row['wdpaid']} not present in DB"
+          next
+        end
+
+       pas_to_update << ProtectedArea.find_by_wdpa_id(row['wdpaid'])
+      end
+      
+      ProtectedArea.where(id: pas_to_update.map(&:id)).update_all(is_dopa: true)
```
Levia, line 28: A couple of considerations here:

1. If you want to make an early check on the existance of the PA, it would better to just save the WDPA ID of the PA found instead of saving the entire PA object and doing a where statement getting the ids again.
2. Further to that, to do way less requests to the DB, you could first get all WDPA IDs from the CSV file and then use the `where` statement to fetch the found PAs. SQL will manage those that can be found automatically.
3. If you are going for the approach above, you could also make a diff between the WDPA IDs in the CSV file and areas found so to then print a statement of the number (and even the related WDPA IDs) of the not found PAs. You could achieve this with something like:

```
  wdpa_ids_from_csv = # read WDPA IDs from CSV
  areas_found = ProtectedArea.where(id: wdpa_ids_from_csv)
  ids_not_found = wdpa_ids_from_csv - areas_found.map(&:wdpa_id)
```

Levia (changes requested):
@stanleypliu Further to the comments I made already, it seems this branch has been branched off of the `refresh-greenlist` one for some reason, which also hasn't been merged into refresh yet. Hence why the extra commits.
This needs to be fixed to avoid issues with the other branch. At this stage, you might need to create another branch and cherry-pick the DOPA related commits from this one and add them to the other one you create

lib/modules/wdpa/dopa_importer.rb
stanleypliu: https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/115#update-63761477 I was under the impression that we had to impose this restriction because of the DOPA Explorer's own limitation.

Levia: I think that already implies that the CSV provided will only contain the WDPA IDs of the areas they show; so those kind of areas are already been excluded at the point of generating the CSV

stacytalbot: This is correct - the whether or not the WDPA ID is present in the CSV is all that needs to be checked.

stanleypliu:
I've created a new branch called `dopa` with these commits, will close this PR and reopen it on that branch once I sort out everything.

Levia:
Sounds good, thanks!

lib/modules/wdpa/dopa_importer.rb
stanleypliu: That's a fantastic idea, also TIL that you can pass an array of IDs to the `where` method.

PR #468, opened 2020-08-06
Ticket 128

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/128


stanleypliu:
@Levia when I initially try to create a layout in the CMS with `{{ cms:categories ... }}` present the `comfy_patching` initializer throws an error regarding the `_layout_category` variable which is nil, but when I add these custom tags to an existing layout it's fine.

stanleypliu:
Found that when I select all three filters, it shows all of my newly created News pages, each of which only have one of the three categories ticked in the CMS, I think it should show only those pages which have all three categories selected? 

Screenshot 2020-08-11 at 10 50 02

stanleypliu:
Just a reminder that we'll need to run `rake cms_categories:import` on the server before everything else to get this working.

PR #469, opened 2020-08-06
Dopa link

stanleypliu:
Ticket

Associated PR for migration: unepwcmc/protectedplanet-db#39

You'll need to run Wdpa::DopaImporter.import in the Rails console to import the CSV, as it's pretty large expect it to take a number of minutes.


Levia:
@stanleypliu What has happened to this one?

stanleypliu:
Rebase didn't work as I thought it would - I need to work out what is going on with `ticket-115` and `dopa`

PR #470, opened 2020-08-06
Ticket 115

stanleypliu:
Ticket

You need to run the migration, then you'll need to run Wdpa::DopaImporter.import in the Rails console to import the CSV, as it's pretty large expect it to take a number of minutes.


stanleypliu:
Method to check whether site is DOPA is there, but the links aren't in the right position nor can I see the link - not sure if I merged something incorrectly:

Screenshot 2020-08-06 at 16 09 37

Levia:
@stacytalbot @stanleypliu I wonder if that has something to do with the merge conflicts?

stanleypliu:
I managed to fix it so that the DOPA link now shows up correctly.

Levia:
Great! Seems to work well. I've just made a very minor change to the way things are logged by the importer. Merging

PR #471, opened 2020-08-06
Ticket 150

Levia:
## Description

* Restore original Pame importer as it wasn't importing the `wdpa_id` field correctly, resulting in some issues also with the download (e.g. showing 0 for that field and having the same evaluation downloaded multiple times because wouldn't be able to match fields using the UNION statement; which statement has also be removed by the way)

* Remove UNION statement for downloading hidden evaluations (where protected_area_id is empty) so that the CSV matches exactly what it shown on the table.

## Notes

[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/150)

### DIFF
--- a/app/models/pame_evaluation.rb
+++ b/app/models/pame_evaluation.rb
@@ -1,5 +1,4 @@
 require 'csv'
-require 'wcmc_components'
 
 class PameEvaluation < ApplicationRecord
   belongs_to :protected_area, optional: true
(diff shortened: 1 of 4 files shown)


app/models/pame_evaluation.rb
```diff
@@ -2,21 +2,12 @@
 require 'wcmc_components'
```
stanleypliu, line 2: Getting issues when running `Wdpa::PameImporter.import`, with or without a file passed, looks like the removal of the wcmc_components gem caused it. Managed to run it successfully with the line removed.

Levia: Ah well spotted! Forgot to remove this line; will do now. Thanks!

stanleypliu:
Works well, I went through the generated CSV to see if I could find any sites in the hidden list, but couldn't.

PR #475, opened 2020-08-11
map styling

WKocur:
1,2,3 and 6 on this ticket: https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/148


WKocur:
@stacytalbot centering icons was a bit tricky - it's pixel rounding issues I guess. This makes it better in my chrome, but a solution fo all browsers looking perfect could be tricky. I could try use an after element instead of background, but that seems less pleasant. let me know if you want me to investigate that.

PR #477, opened 2020-08-12
fixed csv headers

stanleypliu:
CSV headers weren't in the right format for the PAME statistics, so I couldn't import them on the server. I've also changed headers for other files as well to be safe.


lib/data/seeds/pame_country_stats.csv
```diff
@@ -1,4 +1,4 @@
-ISO3,pame_pa_marine_area,pame_pa_land_area,pame_percentage_pa_marine_cover,pame_percentage_pa_land_cover
+"iso3","pame_pa_marine_area","pame_pa_land_area","pame_percentage_pa_marine_cover","pame_percentage_pa_land_cover"
```
lucacug, line 1: the `"` don't make much difference so you can probably leave them, but yeah definitely the ISO3 is an issue.
I would suggest running the import of the stats locally before deploying so you can see issues in advance

stanleypliu: Will update the release manual with that advice, thank you.

PR #478, opened 2020-08-12
Ticket 157

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/157


app/assets/javascripts/comfy/admin/cms/custom.js
```diff
@@ -1,2 +1,5 @@
 //= require 'comfy/admin/cms/versions'
 //= require 'comfy/admin/cms/editor'
+
+console.log('here')
```
stanleypliu, line 4: Shall I remove this?

app/helpers/cms_helper.rb
```diff
@@ -150,4 +150,40 @@ def cta_api
   def cta_live_report
     @cta_live_report ||= CallToAction.find_by_css_class('live-report')
   end
+
+  def get_resource_links 
+    links = []
+
+    if get_resource_link then links.push(get_resource_link) end
+    if get_resource_file then links.push(get_resource_file) end
```
stanleypliu, line 158: I reckon this could be potentially rewritten to combine the `get_resource_links`, `get_resource_link` and `get_resource_file` into one method as I see that there's repeated code.

app/assets/javascripts/comfy/admin/cms/custom.js
stacytalbot: Ah yes sorry

app/helpers/cms_helper.rb
stacytalbot: Yea - my backend Ruby is bad - do you want to refactor?

stanleypliu:
I've pushed a few more commits to try to sanitise the links and provide validations, but there's a problem in that the downloads of attached files don't work properly. I will look into this later.

stanleypliu:
@stacytalbot you should find that the Download link now downloads the file properly instead of it either 404ing or opening the file in a new tab.

PR #481, opened 2020-08-17
Ticket 161

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/161


stanleypliu:
Had a test today of this and made a few dummy resource pages, but the filters still don't work for some reason.

stanleypliu:
It turns out that I was missing the CMS categories tags after my `cms_seeds` folder was wiped when I forgot to back it up one time after I tested my rake task. Now that I put them back in, I can confirm that it's working again.

PR #482, opened 2020-08-18
text replacements in dashboard

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/107

Replaced text in dashboard with Heather's suggestions

### DIFF
--- a/app/helpers/target_dashboard_helper.rb
+++ b/app/helpers/target_dashboard_helper.rb
@@ -1,8 +1,18 @@
 module TargetDashboardHelper
-  def getTooltipText id
+  def getTooltipText(id)
     tooltip = t("thematic_area.target_11_dashboard.tooltips").select { |_tooltip| _tooltip[:id] == id.to_s }.first
+    tooltip ? t("thematic_area.target_11_dashboard.#{id}_text", geo_type: 'global') : ''
+  end
 
-    tooltip ? tooltip[:text] : ''
+  def region_and_country_tooltips
+    t("thematic_area.target_11_dashboard.tooltips").map do |tooltip|
+      id = tooltip[:id]
+      {
+        id: id,
+        title: tooltip[:title],
+        text: t("thematic_area.target_11_dashboard.#{id}_text", geo_type: 'the country/region\'s')
+      }
+    end.to_json
   end
 
   def get_config_carousel_t11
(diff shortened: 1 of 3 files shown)


stanleypliu:
Sorry, I think I committed an older version of the file by accident without realising it should be fixed now

stanleypliu:
I had to basically rework the helpers on that page, as well as the YAML to get two sets of the tooltips up and running without repeating myself.

config/locales/thematic_area/target_11_dashboard/en.yml
```diff
@@ -15,20 +15,25 @@ en:
         -
           id: coverage
           title: Coverage
-          text: "Proportion of the world's terrestrial or marine area covered by protected areas <a href='/thematical-areas/oecms' title='View OECMs' target='_blank'>and OECMs</a>. Statistics updated monthly. Source: <a href='https://www.protectedplanet.net/' title='View source' target'_blank'>WDPA and WD-OECM</a> (UNEP-WCMC & IUCN)"
+          text: &coverage  ! "Proportion of %{geo_type} terrestrial or marine area covered by protected areas <a href='/thematical-areas/oecms' title='View OECMs' target='_blank'>and OECMs</a>. Statistics updated monthly. Source: <a href='https://www.protectedplanet.net/' title='View source' target'_blank'>WDPA and WD-OECM</a> (UNEP-WCMC & IUCN)"
```
stacytalbot, line 18: Remind me what this `&coverage !` does again

app/helpers/target_dashboard_helper.rb
```diff
@@ -1,8 +1,18 @@
 module TargetDashboardHelper
-  def getTooltipText id
+  def getTooltipText(id)
     tooltip = t("thematic_area.target_11_dashboard.tooltips").select { |_tooltip| _tooltip[:id] == id.to_s }.first
+    tooltip ? t("thematic_area.target_11_dashboard.#{id}_text", geo_type: 'global') : ''
```
stacytalbot, line 4: As per Heathers request text should be `the world's` not `global`

config/locales/thematic_area/target_11_dashboard/en.yml
stanleypliu: It is a YAML anchor which allows me to refer and reuse this value elsewhere in the script, which I can access using an alias, which is what the keys with * in front of them are. I used it because I still wanted the tooltips as an array, but wanted a way of referencing them to allow me to interpolate strings into them. The ! is a non-specific tag which means that the value of the YAML node which be resolved to its 'native' data structure

PR #484, opened 2020-08-18
created a blank override to stop categories widget from showing:

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/153


Levia:
@stacytalbot It's probably the easiest and quickest way. @stanleypliu Have you also checked if those categories have been also used elsewhere, like in elasticsearch for searching or other places?

stanleypliu:
No, good point though. I hadn't realised they were indexed in the search.

Levia:
@stanleypliu I'm not sure they are to be honest, but that was just an example to say it would be good to double check if they are getting used around :)

config/locales/search/en.yml
```diff
@@ -36,11 +36,6 @@ en:
     protected-areas: Protected Areas
     results: results
     no-results: No results. Please search again.
-    categories:
```
stacytalbot, line 39: This is actually for the main site search `http://localhost:3000/en/search?search_term=search` and will need putting back.

lib/modules/search.rb
```diff
@@ -2,7 +2,7 @@ class Search
   CONFIGURATION_FILE = File.read(Rails.root.join('config', 'search.yml')).freeze
   ALLOWED_FILTERS = %i(
     type country iucn_category designation region marine has_irreplaceability_info
-    has_parcc_info governance is_green_list category ancestor is_oecm topic page_type
```
stacytalbot, line 5: @Levia i don't know if this is for main site search or not can you check?

lib/modules/search/templates/mappings.json
```diff
@@ -221,17 +221,6 @@
           }
         }
       },
-      "categories": {
```
stacytalbot, line 224: @Levia and this

PR #491, opened 2020-08-24
Allow custom datetime tag to work

stanleypliu:
Bug fix for staging, to allow the custom tag `date_not_null` to be properly sorted in the views as the Comfy helper `cms_fragment_content` is not properly configured to work with it (returns its content which is nil).

### DIFF
--- a/app/helpers/cms_helper.rb
+++ b/app/helpers/cms_helper.rb
@@ -194,9 +194,7 @@ module CmsHelper
   end
 
   def cms_fragment_content_datetime(identifier, page)
-    return cms_fragment_content(identifier, page) unless identifier == 'published_date'
-    date_fragments = page.fragments.where(identifier: 'published_date')
-    date_not_null = date_fragments.find_by(tag: 'date_not_null')
-    date_not_null ? date_not_null.datetime : date_fragments.first.datetime
+    date_not_null = page.fragments.find_by(tag: 'date_not_null', identifier: identifier)
+    date_not_null.datetime 
   end
 end
--- a/app/helpers/cms_helper.rb
+++ b/app/helpers/cms_helper.rb
@@ -194,6 +194,7 @@ module CmsHelper
   end
 
   def cms_fragment_content_datetime(identifier, page)
-    page.fragments.find_by(tag: 'date_not_null', identifier: identifier).datetime
+    # Might not be able to find that particular tag and identifier combo -  hence the try
+    page.fragments.find_by(tag: 'date_not_null', identifier: identifier).try(:datetime)
   end
 end
--- a/app/helpers/cms_helper.rb
+++ b/app/helpers/cms_helper.rb
@@ -193,8 +193,10 @@ module CmsHelper
     end
   end
 
-  def cms_fragment_content_datetime(identifier, page = @cms_page)
-    return cms_fragment_content(identifier, page) unless page.fragments.pluck(:tag).include?('date_not_null')
-    page.fragments.find_by(tag: 'date_not_null').datetime
+  def cms_fragment_content_datetime(identifier, page)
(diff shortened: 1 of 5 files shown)


app/helpers/cms_helper.rb
```diff
@@ -192,4 +192,8 @@ def get_resource(fragment_text, fragment_link, button_text, button_class)
       false
     end
   end
+
+  def cms_fragment_content_datetime(identifier, page)
+    page.fragments.find_by(tag: 'date_not_null', identifier: identifier).datetime
```
Levia, line 197: @stanleypliu When possible, it's always a good idea to prevent possible exceptions and return a fallback value.
In this case for example, even though it's not very likely, it might happen that you can't find a fragment with that tag and identifier; so if that find returns `nil` your code will throw an exception with `datetime` because it can't me invoked on nil. So you could make use of `try`, like: `find_by(...).try(:datetime)` which invokes the method if the previous result is not nil, returns nil otherwise.

app/serializers/search/cms_serializer.rb
```diff
@@ -57,7 +57,7 @@ def sorted_pages
   end
 
   def date(page)
-    _date = cms_fragment_content(:published_date, page)
+    _date = cms_fragment_content_datetime(:published_date, page)
```
stanleypliu, line 60: Thanks for the comment Ferdi. There's a somewhat related point here about using `cms_fragment_content_datetime` in the CMS serializer - currently all the datetime fragments in my CMS layouts are `date_not_null` tags but there's the possibility that it might not be sometimes. In this case, would we want to continue returning `_date` even if it's `nil` (which is what the tertiary condition is checking for) ?

Levia: @stanleypliu Good point. I think it might be useful to return check of the default date tag in case date_not_null is not used. Which is something you somewhat did previously and then removed from the content_datetime method :)

So perhaps your custom method could be something like:

```
 def cms_fragment_content_datetime(identifier, page)
    date = page.fragments.find_by(tag: 'date_not_null', identifier: identifier).try(:datetime)
    date ? date : page.fragments.find_by(tag: 'date', identifier: identifier).try(:datetime)
  end
```

So the above is very specific for dates kind of fragments. Otherwise, it could also be:

```
 def cms_fragment_content_datetime(identifier, page)
    date = page.fragments.find_by(tag: 'date_not_null', identifier: identifier).try(:datetime)
    date ? date : cms_fragment_content(identifier, page)
 end
```

which is more general, but if misused it could return the content of another tag, like a normal content string, which could make the strftime line below fail.

What do you think?

stanleypliu: Second method makes more sense to me - I originally intended this method to fall back to using the Comfy helper if it was passed an non-date tag.

Levia: Yeah, if you are going for the second method I've written above that is a good idea!
Let me know when this is ready to be reviewed again :)

app/helpers/cms_helper.rb
```diff
@@ -192,4 +192,10 @@ def get_resource(fragment_text, fragment_link, button_text, button_class)
       false
     end
   end
+
+  def cms_fragment_content_datetime(identifier, page)
+    # Might not be able to find that particular tag and identifier combo -  hence the try
+    page.fragments.find_by(tag: 'date_not_null', identifier: identifier).try(:datetime)
```
Levia, line 198: @stanleypliu I think you forgot to assign this value to the `date` var here :)

Levia: Isn't it throwing an exception for you?

stanleypliu: 😓

Levia:
Great! Merging this now!

PR #493, opened 2020-08-24
Ticket 171

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/171

### DIFF
--- a/app/javascript/components/charts/am-chart-pie/AmChartPie.vue
+++ b/app/javascript/components/charts/am-chart-pie/AmChartPie.vue
@@ -1,0 +1,78 @@
+<template>
+  <div class="am-chart--pie">
+    <div class="chart__chart">
+      <div 
+        class="chart__svg"
+        id="chart-pie"
+      />
+    </div>
+  </div>
+</template>
+
+<script>
+import * as am4core from "@amcharts/amcharts4/core"
+import * as am4charts from "@amcharts/amcharts4/charts"
+
+export default {
+  name: 'AmChartPie',
+
+  props: {
+    dataset: {
+      required: true,
+      type: Array // [{ title: String, value: Number }]
+    },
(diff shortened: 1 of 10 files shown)


app/javascript/components/charts/am-chart-pie/AmChartPie.vue
```diff
@@ -0,0 +1,94 @@
...
+        am4core.color('#5F81CB'),
+        am4core.color('#FAA51B'),
+        am4core.color('#EF5F6C'),
+        am4core.color('#151617'),
+        am4core.color('#71A22B'),
+        am4core.color('#F5F58A'),
+        am4core.color('#EF266C'),
+        am4core.color('#1A4D9F'),
+        am4core.color('#E57133')
+      ]
```
stanleypliu, line 75: I was wondering whether you could define this list in a mixin or put it in the global scope somehow if it's going to be constant, then I came across this which was quite interesting: https://til.hashrocket.com/posts/sxbrscjuqu-share-scss-variables-with-javascript

stanleypliu:
My legend doesn't have colours 
Screenshot 2020-08-24 at 15 05 30
Screenshot 2020-08-24 at 15 07 01

app/javascript/components/charts/am-chart-pie/AmChartPie.vue
stacytalbot: Ah yea - atm only the pie chart is using this colour scheme.  I think when we get to the point where the scss is being handled by webpack we should def look into sharing the variables from the css to the js.

I will tidy up the method a bit when i do the governance chart - I just wanted to get it built so that charlotte can check functionality.

stanleypliu:
It's showing now, looks good and the colours on the chart are more distinct now as well.

PR #494, opened 2020-08-25
Sources table

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/170

Added data for the sources table in the form: 

```
[ { title: ..., responsible_party: ..., year: ... } ... ]
```

Awaiting rework from Charlotte as the 'Details' link is no longer a link to a URL, but will be the party that created the source.

### DIFF
--- a/app/assets/stylesheets/components/_lists.scss
+++ b/app/assets/stylesheets/components/_lists.scss
@@ -24,7 +24,7 @@
     &__title {
       font-weight: $bold;
       margin-right: rem-calc(10);
-    } 
+    }
 
     &__right {
       margin-left: auto;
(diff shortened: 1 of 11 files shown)


Levia:
@stanleypliu I think it would be worth thinking about an SQL query to retrieve the sources data.
With your current solution, you are hitting the database for each PA, which can be computationally expensive and troublesome for the database.

You could probably take inspiration from other methods in the country and region models like `protected_areas_per_governance` and especially from the `protected_areas_inner_join` one

stanleypliu:
The computational cost ramps up massively for regions, it's not so noticeable for the country and PA pages though.

Levia:
It's not specifically on how much time it requires to fetch the data, but more on the number of times the database gets hit.
Hitting the database N number of times for each item is not recommandable for potentially large numbers and especially if you could get away with hitting it just once. Database sessions can be a pain :)

PR #495, opened 2020-08-26
Ticket 125

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/125


Levia:
@stacytalbot Seems to work well now. Thanks!
I've also ensured that empty queries are not ignored now and merged the latest from refresh.
Feel free to merge if you are happy with it. I'll update the related ticket

PR #498, opened 2020-08-26
Populate links

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/179

Links have been populated on the homepage with the exception of the API CTA which will be updated manually on staging.


stanleypliu:
@stacytalbot Apologies about that, the links in the footer should now link to the social media accounts, whilst the rest of the site, as far as I could tell, uses the share links instead. Also, I've changed the CTA to use the API URL. 

Screenshot 2020-08-27 at 17 19 22

PR #499, opened 2020-08-26
Ticket 180

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/180


stanleypliu:
Footer looks better. Was there any plan to add a message to the search on the map if no match was found (alongside the change in background colour)?

PR #501, opened 2020-08-27
Ticket 171 links

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/171

Populated the IUCN categories chart links with search results relating to the various categories for each country/region's PAs.

### DIFF
--- a/app/helpers/countries_helper.rb
+++ b/app/helpers/countries_helper.rb
@@ -7,6 +7,24 @@ module CountriesHelper
     @country.children.any? || @country.parent.present?
   end
 
+  def iucn_link(category)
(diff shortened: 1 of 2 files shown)


Levia:
@stanleypliu This seems to work well. Although I think we need to add the same for the governances links

stanleypliu:
I will add that on `ticket-33`

Levia:
@stanleypliu What's the reason for that? It feels more closely related to this work than to the downloads one

Levia:
@stacytalbot Ah I see now. Thanks!

PR #502, opened 2020-08-27
Sync seeds adjustments

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/177

Wasn't just adjustments in the end, I ended up having to fix the `commence_import` method as it wasn't importing correctly (wrong source). Was probably a good idea that I left it commented out!

### DIFF
--- a/lib/modules/sync_seeds.rb
+++ b/lib/modules/sync_seeds.rb
@@ -140,7 +140,7 @@ class SyncSeeds
   def commence_comfy_import(answer, source)
     logger = ComfortableMexicanSofa.logger
     ComfortableMexicanSofa.logger = Logger.new(STDOUT)
-    site = Comfy::Cms::Site.first
+    site = Comfy::Cms::Site.first.identifier
     folder = source.split('/').last
 
     if answer == 'all'
--- a/lib/modules/sync_seeds.rb
+++ b/lib/modules/sync_seeds.rb
@@ -140,7 +140,7 @@ class SyncSeeds
   def commence_comfy_import(answer, source)
     logger = ComfortableMexicanSofa.logger
     ComfortableMexicanSofa.logger = Logger.new(STDOUT)
-    site = Comfy::Cms::Site.find_by_locale(I18n.locale).identifier
+    site = Comfy::Cms::Site.first
     folder = source.split('/').last
 
     if answer == 'all'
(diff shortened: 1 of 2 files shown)


lib/modules/sync_seeds.rb
```diff
@@ -137,18 +137,17 @@ def main_task(local_list:, remote_list:, local_base:, remote_base:)
...
 
   # Piggybacks on existing Comfy modules 
-  # TODO - the arguments used here for the site and folder
-  #  (e.g. protectedplanet or protected-planet) can be different 
-  # for different environments, need to change both this and the staging seeds rake task
-  def commence_comfy_import(answer)
+  def commence_comfy_import(answer, source)
     logger = ComfortableMexicanSofa.logger
     ComfortableMexicanSofa.logger = Logger.new(STDOUT)
+    site = Comfy::Cms::Site.find_by_locale(I18n.locale).identifier
```
Levia, line 143: I think that finding the site by locale is not exactly what we originally intended and also not sure it would always return the expected results.
The identifier of a site is, indeed, the `identifier` attribute, so what originally had in mind is to pass a site identifier to the `commence_comfy_import` method so that it could also fetch the desired site by that; and, if not found, it would return a properly formatted error.
How does that sound?

stanleypliu: It is true that it might not always return the expected results, but we only have one locale at the moment, and it doesn't look like there's plans to add any more for the time being. If I understand correctly we'll only be using this a limited number of times, once if everything goes to plan, so I wasn't sure whether this was necessary. Opting for `find_by_locale` was easier in my mind, but if you want me to change it, I can include that `find_by_locale` call as the fallback if no identifier is found to match what is in the list of sites, and if that doesn't work, then I can raise an error

Levia: Yeah I mean, at this stage it's not strictly necessary to add that bit really, it was mainly for testing if we wanted to test things either locally or staging before doing anything on prod. 

At the same time, finding the site by locale doesn't sound great for the same considerations. Given there's going to be only one site for the time being you could have just fetched with `Comfy::Cms::Site.first` I suppose

lib/modules/sync_seeds.rb
```diff
@@ -140,7 +140,7 @@ def main_task(local_list:, remote_list:, local_base:, remote_base:)
   def commence_comfy_import(answer, source)
     logger = ComfortableMexicanSofa.logger
     ComfortableMexicanSofa.logger = Logger.new(STDOUT)
-    site = Comfy::Cms::Site.find_by_locale(I18n.locale).identifier
+    site = Comfy::Cms::Site.first
```
Levia, line 143: You should get the identifier though, right?

PR #503, opened 2020-08-27
Return empty string if attachment is missing

Levia:
## Description

Fix 500 error when clicking on resources pages.
The error seems related to missing attachments still. Cleaning up the staging db more might help as well.
Here I'm returning an empty string instead if attachments are missing, which should be more future proof.

### DIFF
--- a/app/helpers/cms_helper.rb
+++ b/app/helpers/cms_helper.rb
@@ -167,7 +167,8 @@ module CmsHelper
   def linkify(url, fragment_link)
     # If it is a file as opposed to a link
     if fragment_link =~ /(file)/ 
-      file = @cms_page.fragments.find { |fragment| fragment.tag == 'file' }
+      file = @cms_page.fragments.find_by(tag: 'file')
+      return '' if file.attachments.first.blank?
       return rails_blob_path(file.attachments.first, disposition: 'attachment')
     end
     # TODO - add validations to Comfy pages controller to make it more robust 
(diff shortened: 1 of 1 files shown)


app/helpers/cms_helper.rb
```diff
@@ -167,7 +167,8 @@ def get_resource_links
   def linkify(url, fragment_link)
     # If it is a file as opposed to a link
     if fragment_link =~ /(file)/ 
-      file = @cms_page.fragments.find { |fragment| fragment.tag == 'file' }
+      file = @cms_page.fragments.find_by(tag: 'file')
+      return '' if file.attachments.first.blank?
```
stanleypliu, line 171: Still receiving an issue with this on certain pages, like http://localhost:3000/resources/august-2019-update-of-the-wdpa - specifically a no method error as `file` is nil. I did have a closer look at this, and I found that even though the `fragment_link` was `:resource_file`, the tag was `text`, not `file`! However, this may be an unrelated issue that is most likely caused by the seed import.

stanleypliu:
Just tried this with a test resource, it works well even though the resulting URL is blank. Perhaps a better idea would be to disable the link altogether if we don't have a file/URL?

Levia:
Thanks @stanleypliu . I've added a quick fix so it check if the file exists as well.
With regards disabling the link that might be a good idea too, perhaps something for another ticket or for later though; so we can have this merged and deployed to staging asap. Not sure if @stacytalbot agrees

stanleypliu:
I'm happy to merge then if everything else is fine.

stanleypliu:
Just added a commit of my own to deal with the old-style resource pages not displaying properly because of nil attributes of the `get_resource_links` method.

PR #505, opened 2020-08-28
Remove types filter

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/187

Quick fix to remove Topics filters from News and Stories index page.

### DIFF
--- a/config/initializers/comfy_patching.rb
+++ b/config/initializers/comfy_patching.rb
@@ -36,10 +36,8 @@ Rails.configuration.to_prepare do
       
       own_layouts_categories = filter_layouts_categories(_categories)
 
-      unless own_layouts_categories.blank?
-        delete_orphan_categories(own_layouts_categories) 
-        return if _categories.blank?
-      end
+      delete_orphan_categories(own_layouts_categories) unless own_layouts_categories.blank?
+      
 
       _categories.each do |cat|
         tag_name = cat[:tag_params]
--- a/config/initializers/comfy_patching.rb
+++ b/config/initializers/comfy_patching.rb
@@ -35,10 +35,10 @@ Rails.configuration.to_prepare do
       # Not handled at present - adding new categories which equates to adding new layout categories which means _layout_category is empty
       
       # layouts_categories is in the format [{identifier:... label: 'topics'...}, {identifier:..., label: 'types'...}]
-      own_layouts_categories = Comfy::Cms::LayoutsCategory.where(layout_id: self.id)
+      own_layouts_categories = Comfy::Cms::LayoutsCategory.where(layout_id: self.id).to_a
 
       _categories.each do |_cat| 
-        own_layouts_categories.to_a.filter! { |category| _cat[:tag_params] != category.layout_category.label }
+        own_layouts_categories.filter! { |category| _cat[:tag_params] != category.layout_category.label }
       end
 
       delete_orphan_categories(own_layouts_categories) unless own_layouts_categories.blank?
--- a/config/initializers/comfy_patching.rb
+++ b/config/initializers/comfy_patching.rb
@@ -56,7 +56,7 @@ Rails.configuration.to_prepare do
     def delete_orphan_categories(own_layouts_categories)
       self.pages.each do |page|
         own_layouts_categories.each do |lc|
-          pcs_ids = page.page_categories.where(layout_category_id: lc.id).map(&:id)
+          pcs_ids = page.page_categories.where(layout_category_id: lc.layout_category_id).map(&:id)
           Comfy::Cms::PagesCategory.where(
             page_id: page.id,
             page_category_id: pcs_ids
--- a/config/initializers/comfy_patching.rb
+++ b/config/initializers/comfy_patching.rb
@@ -21,6 +21,7 @@ Rails.configuration.to_prepare do
     after_save :assign_layout_categories
 
     def assign_layout_categories
+      # _categories = [{ tag_params: 'topics', tag_class: 'categories'}]
       _categories = self.content_tokens.select do |t|
         unless t.is_a?(Hash)
           false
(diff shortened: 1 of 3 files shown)


Levia:
@stanleypliu I think this should have been a matter of making sure that the those categories have been removed from the layout successfully and, also, that upon updating the layout the related database objects are also deleted correctly (I thought this was managed already by the comfy patching, perhaps not).
Please do let me know if I missed something with regards the desired intent of the ticket

stanleypliu:
If they are to be deleted from the database upon removing the tag from the layout, I'm not sure that that's currently handled by the Comfy patcher - we would need to add a callback to the Comfy layout controller though to do that I feel. I could also look at putting that in.

Levia:
@stanleypliu Yes please do. If I remember correctly that after_save method should also destroy the related `Comfy::Cms::LayoutsCategory` objects; and I believe it's not doing so at the moment

stanleypliu:
Give it a try now - have also renamed a few of the `layout_categories` to `layouts_categories`

config/initializers/comfy_patching.rb
```diff
@@ -29,7 +29,10 @@ def assign_layout_categories
         end
       end
 
-      delete_orphan_categories && return if _categories.blank?
+      layouts_categories = Comfy::Cms::LayoutsCategory.where(layout_id: self.id)
+      orphan_categories = layouts_categories.to_a - _categories
```
Levia, line 33: @stanleypliu Not sure what this is supposed to do?
`layouts_categories.to_a` is an array containing `LayoutsCategory` objects, while `_categories` is an array of hashes with information about the layout and its tags. So theoretically this operation always returns `layouts_categories.to_a` regardless of what's inside `_categories`

stanleypliu: You're right - I mistook the `_categories` variable to also be an array of LayoutsCategory objects which were present before the save, didn't pick up on this because it correctly deletes the categories if they aren't present.

stanleypliu: so i've dispensed with subtracting `_categories` from `layouts_categories` and just settled for filtering `layouts_categories` to get the orphans. one thing i'm not too sure about is how you register new categories tags, it doesn't seem to be as simple as adding the new label after the tag in the CMS editor. If we know we are just going to have topics and types as the two main category types, should we have error handling in case anyone editing the CMS in the future tries to add a new custom label?

config/initializers/comfy_patching.rb
```diff
@@ -29,7 +29,19 @@ def assign_layout_categories
...
 
-      delete_orphan_categories && return if _categories.blank?
+      # Two situations for deleting orphan categories, either _categories has no or a restricted set of the currently existing categories
+      # Or, _categories consists of tags with entirely new labels, and hence new layouts would have to be created
+      # Not handled at present - adding new categories which equates to adding new layout categories which means _layout_category is empty
+      # _categories = [{ tag_params: 'topics', tag_class: 'categories'}]
+      # layouts_categories = [{label: 'topics'...}, {label: 'types'...}]
+      # orphan_categories = [{label: 'types'}]
+      layouts_categories = Comfy::Cms::LayoutsCategory.where(layout_id: self.id)
+      .map(&:layout_category)
```
Levia, line 39: I would add this in the same line above or add some indentation here (2 spaces) so to understand that it's something referring to a preceding code

config/initializers/comfy_patching.rb
```diff
@@ -29,7 +29,19 @@ def assign_layout_categories
...
+      _categories.each do |_cat| 
+        layouts_categories.filter! { |category| _cat[:tag_params] != category.label }
```
Levia, line 41: Care this is not actually deleting layouts categories. This means that even if you end up with a `layouts_categories` array of one element only, the `delete_orphan_categories` is still fetching all `layouts_categories` of the given layout regardless.

Example. If you have 2 layouts categories and remove one, `delete_orphan_categories` will remove both anyway

stanleypliu: It is deleting the filtered `layouts_categories`, not the original `layouts_categories` that is defined in line 38.

Levia: Line 67 in the `delete_orphan_categories` is fetching all layouts_categories of the layout, not the local variable defined in this method, exactly because `layouts_categories` in line 38 is a local variable only visible within `assign_layout_categories`.

So line 67 is basically doing `self.layouts_categories.map(&:destroy)` , which can also be seen as `Comfy::Cms::LayoutsCategory.where(layout_id: self.id).map(&:destroy)`

Perhaps it would also be good to rename the local variable on line 38 to avoid confusion?

Please give this a test again on your local environment yourself to ensure this is actually how it's behaving.
Given 2 categories and deleting 1, it results in the `Comfy::Cms::LayoutsCategory.where(layout_id: the_layout_id)` to be empty for me.

stanleypliu: Sorry, you're right, the way I evaluated this was a simple eye test to see whether the Types filters were removed from the News and Stories page if I deleted the types tag, and they were. I now realise that there's more to it especially if we were to export and import the updated News and Stories layout

stanleypliu: OK, I believe it was because it was not creating the new LayoutsCategory instances after it deleted the categories as well - it was just immediately returning and not proceeding to the `find_or_create_by` loop

Levia:
@stanleypliu I fixed that a bit more.
First of all, doing `your_var.to_a.filter!` is not going to overwrite `your_var` as the array created with `to_a` is generating an entirely new object, so `filter!` is acting on that new object instead. To fix it I assigned the `to_a` version of the relationship directly into the variable.

Secondly, I found that pages_categories objects were not deleted. That is because we were trying to fetch those objects comparing `layout_category_id` with at `layouts_category_id` instead (note the extra 's'). That is probably my bad with naming these tables like so!

Could you please test this on your end again before merging?

stanleypliu:
Thank you for resolving the confusion with the page categories (or pages categories). I think the results I've been seeing have all been false positives, since the LayoutsCategory collection for the News and Stories template was being correctly updated with every change I was making, so I wasn't able to pick up on errors in my code as easily. 

When testing, I was able to get the LayoutsCategory collection to update for the news template but not for the resources template until I restarted the server and was able to see that updating as well.

stanleypliu:
@Levia I've added extra support for the event that `_categories` is blank: I'm not sure how it's not broken once it reaches the loop that regenerates the LayoutsCategory objects for the layout. Anyway, it'll return if the array is empty

PR #507, opened 2020-08-28
Rework helper for resource links

Levia:
## Description

Rework helper for resource links.
This should fix the behaviour reported in the related ticket:

```
If a link or file are not provided then they should not be shown.
```

Needs to be properly tested on staging or by getting the CMS seeds from staging.


app/helpers/cms_helper.rb
```diff
@@ -154,45 +154,41 @@ def cta_live_report
...
+    if fragment_link.tag == 'file'
+      if Rails.env.development?
+        return rails_blob_path(fragment_link.attachments.first)
+      else
+        return fragment_link.attachments.first.service_url&.split('?')&.first
+      end
     else
-      nil
+      url = fragment_link.content
+      url =~ /(http:|https:)/ ? url : 'https://' + url
```
stanleypliu, line 191: Should we continue to have this line in? This was only a temporary fix to get round the issue of valid URLs, I feel that if you are going to make a change to the methods, then this would also be a good opportunity to make this a more robust fix 😉

Levia: Thanks, I've updated this to double check the URL string and return an 'Invalid' string if not valid.
Could you and/or @stacytalbot double check that the partial related to this helper is actually being used and where? I found it difficult to test this properly as I get confused between the `partials/cards/_resources.html.erb` partial (where links and files are showing properly although using different CMS tags to fetch those) and the `layouts/cms/partia/_resource.html.erb` where differently named tags are meant to be used.

stanleypliu: That is the `layouts/cms/partial/_resource.html.erb` which concerns individual resource pages, and is used for them - not the `partials/cards/_resources.html.erb` which you can think of as the index view of all the resources present in the CMS.

Levia: Thanks for claryfing!
So the question (perhaps more for @stacytalbot ) is whether files and links in the individual resource pages should be using different CMS tags (perhaps referring to different things) than the tags used for the home page?

stacytalbot: Yes they are different.

A resource can have a link instead of a show page. 
A resource can have a file instead of a show page.
A resource page can have a show page which can have a link and a file.

stacytalbot: Resource with a link.
Resource with a file.
Resource with a show page.

Screenshot 2020-09-01 at 12 34 29

stacytalbot: Resource with a show file that has a link and a file (in future multiple links and files will be allowed to be added to this page)

Screenshot 2020-09-01 at 12 36 51

stacytalbot: Is that clear? Or do you need me to explain further?

app/helpers/cms_helper.rb
```diff
@@ -154,45 +154,41 @@ def cta_live_report
...
+  def linkify(fragment_link)
+    if fragment_link.tag == 'file'
+      if Rails.env.development?
+        return rails_blob_path(fragment_link.attachments.first)
+      else
+        return fragment_link.attachments.first.service_url&.split('?')&.first
+      end
     else
-      nil
+      url = URI.parse(fragment_link.content)
```
stanleypliu, line 190: It fails on this line if `fragment_link.content` is not a URI - for instance, the global reports resource has a '\n' as its content. Might need a begin-rescue block or something to catch the error.

app/helpers/cms_helper.rb
Levia: Perfectly clear now, thanks!

app/helpers/cms_helper.rb
Levia: Well spotted. Fixed it now

app/helpers/cms_helper.rb
```diff
@@ -154,45 +154,45 @@ def cta_live_report
...
+      if Rails.env.development?
+        return rails_blob_path(fragment_link.attachments.first)
+      else
+        return fragment_link.attachments.first.service_url&.split('?')&.first
+      end
     else
-      nil
+      begin
+        url = URI.parse(fragment_link.content)
+        (url.is_a?(URI::HTTP) && !url.host.nil?) ? url.to_s : 'Invalid or missing URL'
```
stanleypliu, line 192: It's also appending these messages to the root URL as part of the generated URL - will need to have a think.

Levia: So the intent here was to return something that is not a URL (so a normal string) so that it won't create, possibly, any link. Perhaps what you are mentioned is something to be done on the frontend and ideally change those strings into a default value that could be more easily parsed by the frontend itself?

stanleypliu: I've made a couple of changes to the helper and put in a condition in the partial to only display the list item with the link within if the URL is not blank.

stanleypliu: If this approach is sound, I can merge this in.

app/helpers/cms_helper.rb
```diff
@@ -181,6 +181,7 @@ def get_resource(fragment_text, fragment_link, button_text, button_class)
   # Also sanitises it in the case of a download link
   def linkify(fragment_link)
     if fragment_link.tag == 'file'
+      return '' if !fragment_link.attachments || fragment_link.attachments.first.blank?
```
Levia, line 184: It would be better to be consistent with what we return in case of missing links or attachments. So I would suggest `nil` both here and below.

app/views/layouts/cms/_resource.html.erb
```diff
@@ -15,10 +15,12 @@
     <% unless get_resource_links.empty? %>
       <ul class="list--links">
         <% get_resource_links.each do |link| %>
-          <li class="list__li">
-            <span><%= link[:text] %></span>
-            <%= link_to link[:button], link[:url], title: link[:text], class: "list__a #{link[:classes]}", target: '_blank' %>
-          </li>
+          <% unless link[:url].nil? || link[:url].blank? %>
```
Levia, line 18: `.blank?` already accounts for nil values as well, so the first bit can be omitted

app/helpers/cms_helper.rb
stanleypliu: OK, I have made the changes

PR #508, opened 2020-09-01
populating coverage on homepage

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/39

Added % PA coverage figure for homepage

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -4,7 +4,7 @@ class HomeController < ApplicationController
   def index
     @presenter = HomePresenter.new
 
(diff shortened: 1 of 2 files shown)


Levia:
@stanleypliu I've update the calculations of global land and marine cover slightly, as I think it's best to calculate those with the already existing stats rather than redoing the calculations based on the percentages.
Happy for this to be merged if you are happy

stanleypliu:
I seem to recall that you said the CountryStatistic data hadn't been updated in a while - then again I'm using a value from that model for the methods. I don't mind in all honesty - the important thing is that the coverage value is properly calculated. I'm going to merge this.

PR #509, opened 2020-09-01
Ticket 63

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/63


stanleypliu:
I tested this with a few sites, but it doesn't show the maps still 
Screenshot 2020-09-01 at 15 37 03
Screenshot 2020-09-01 at 15 35 53

stanleypliu:
Also, the maps don't load for the site/country/region pages, but on `refresh` they do. Is that the map loading issue you mentioned?

stanleypliu:
Maps are working - I will merge after ticket 33.

PR #510, opened 2020-09-01
use old pp maps

WKocur:
This hopefully will work - I couldn't test locally. It uses the old tilesets (which are added into a style). This does mean that the basemap is no longer a series of layers, meaning polygons will have to be added on top of the place names and boundaries, rather than beneath them.


WKocur:
@stacytalbot Possibly a race condition - hard to debug without having this running locally... I can pair program it with someone running locally? It's trying to add the layer twice for some reason...

WKocur:
Alternatively can come back to this once tradeview is sorted and I don't mind reinstalling postgres - it should cause issues beyond the error in the console.

PR #500, opened 2020-08-26
Ticket 108 

lucacug:
This is to refactor the target11 api module to accomodate the new developed jrc endpoints.

:warning: please ask for the most uptodate .env file before testing/deploying
:warning:  :warning:  To correctly populate the dashboard you would need to import the new oecm csv via `ImportTools.statistics_monthly_import`

This also includes the [ticket 189](https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/189)

Relative DB [PR](https://github.com/unepwcmc/protectedplanet-db/pull/40)

PR #514, opened 2020-09-03
ticket 196

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/196


lib/modules/stats/global.rb
```diff
@@ -42,6 +42,11 @@ def self.calculate_stats_for(klass, field_name)
     end
   end
 
+  def self.percentage_oecm_coverage
+    ((ProtectedArea.global_terrestrial_oecm_coverage 
+    + ProtectedArea.global_marine_oecm_coverage) / self.global_area).to_f * 100
```
Levia, line 47: These are probably going to be reworked with the changes made by Luca in PR #500 . Ed has provided the OECM country stats which should be preferred for this kind of calculations

stanleypliu: Had a brief look at the new methods in the Stats::Global module - I will change to use those methods when that PR's been merged.

Levia: I'm double checking how it's best to calculate the OECM only stats as there are multiple ways we could do that, but no idea which one should we use. I'll let you know when I get an answer

Levia:
@stacytalbot @stanleypliu This is ready now

PR #515, opened 2020-09-03
remove global from list of regions on search results page

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/197

Remove global region from list of regions on search results page


Levia:
@stanleypliu I changed this so that the region is not indexed to begin with. This ensures that won't appear in any search or autocompletion. Hope it makes sense. Please merge if happy

PR #516, opened 2020-09-03
Ensure OECM attributes can be set only if PA is an OECM area

Levia:
## Description

Amend model and importer to account for the new OECM attributes stored in the ProtectedAreas table

Related [DB PR](https://github.com/unepwcmc/protectedplanet-db/pull/41)

### DIFF
--- a/app/models/protected_area.rb
+++ b/app/models/protected_area.rb
@@ -24,6 +24,8 @@ class ProtectedArea < ApplicationRecord
 
   after_create :create_slug
 
+  validate :oecm_attributes
+
   scope :all_except, -> (pa) { where.not(id: pa) }
 
   scope :oecms, -> { where(is_oecm: true) }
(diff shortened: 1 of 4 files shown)


Levia:
@lucacug Would you please be able to have a quick look at this one?
I believe it might be quicker for you to understand the changes here

app/models/protected_area.rb
```diff
@@ -326,7 +328,10 @@ def self.with_valid_iucn_categories
     )
   end
 
+  OECM_ATTRS_ERROR = "'conservation_objectives' and 'supplementary_info can only be assigned if this is an OECM area".freeze
```
lucacug, line 331: I think there may be a `'` missing after supplementary_info?

Levia: Well spotted! Fixed now. Thanks!

PR #518, opened 2020-09-04
fixed polygon and points count on region pages

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/200

Fixed polygons-points count on region pages

### DIFF
--- a/app/presenters/region_presenter.rb
+++ b/app/presenters/region_presenter.rb
@@ -58,13 +58,14 @@ class RegionPresenter
   def geometry_ratio
     polygons_count = 0
     points_count = 0
-    total = polygons_count + points_count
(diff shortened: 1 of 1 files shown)


app/presenters/region_presenter.rb
```diff
@@ -58,14 +58,14 @@ def total_polygons_percentage
   def geometry_ratio
     polygons_count = 0
     points_count = 0
-    total = 0
-
+    total = polygons_count + points_count
```
Levia, line 61: Isn't that going to stay `0` anyway? at the moment of assigning the value to `total` the other 2 variables are 0

PR #521, opened 2020-09-07
added new green list data from september release

stanleypliu:
Added the new statistics from the September WDPA release 

Need to run `Wdpa::GreenListImporter.import_global_data` on staging to see the changes

### DIFF
--- a/lib/data/seeds/green_list_data.csv
+++ b/lib/data/seeds/green_list_data.csv
@@ -1,0 +1,4 @@
+type,value
+green_list_perc,0.14
+green_list_area,726072.3143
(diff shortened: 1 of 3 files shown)


Levia:
@stanleypliu That's more a question for the data provider, but 38 seems quite low considering that the number of GL sites we are given every month is >= 102.

stanleypliu:
Ed's confirmed that it's fine to show the absolute values as per the data on the GL page, and if IUCN take issue then we would change it, but not until they do.

Levia:
Thanks @stanleypliu ! Merging

PR #523, opened 2020-09-15
Ticket 203

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/203

PR for migration: https://github.com/unepwcmc/protectedplanet-db/pull/42

I've added `update_year` (corresponding to the `update_yr` column in the standard_sources table of the WDPA) to the sources data standard module, as well as correcting the class methods responsible for populating the sources tables so that they now use the correct attribute. 

@Levia Please let me know if this is all that is needed - I've checked myself and the Source model seems well isolated from the rest of the database. 

We'll need to re-import the WDPA on staging, or just the sources if it's possible to do so to see the changes.

### DIFF
--- a/app/models/country.rb
+++ b/app/models/country.rb
@@ -73,7 +73,7 @@ class Country < ApplicationRecord
 
   def sources_per_country
     sources = ActiveRecord::Base.connection.execute("""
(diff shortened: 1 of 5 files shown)


Levia:
@stanleypliu Well done! This looks good to me.
Merging

PR #524, opened 2020-09-15
Ticket 204

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/204

Added a min height of the `filtered-table` component on tablet upwards so that the country filter modal doesn't get obscured by the footer.

### DIFF
--- a/app/assets/stylesheets/components/_table.scss
+++ b/app/assets/stylesheets/components/_table.scss
@@ -42,11 +42,7 @@ $table-horizontal-scroll-row-width-desktop: 80vw;
 // classes
 //--------------------------------------------------
 .filtered-table {
-  min-height: $filters-menu-max-height + rem-calc(80);
-
-  @include breakpoint($small) {
-    min-height: $filters-menu-max-height + rem-calc(200);
-  }
+  min-height: $filters-menu-max-height + rem-calc(200);
 }
 
 //--------------------------------------------------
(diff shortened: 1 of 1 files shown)


Levia:
@stanleypliu This works well for desktop and tablet, but the issue still appears for mobile. Are we going to fix that too?

stanleypliu:
Would you mind showing a screenshot to better describe what you are talking about? The selection box for the countries takes up the whole screen when I looked on Google Dev Tools, so I thought that it wouldn't be an issue - this is what it looks like for me on 360px, the lowest supported resolution. I know the buttons are touching the bottom, but the footer isn't present anyway: 

Screenshot 2020-09-16 at 16 47 18

Levia:
Sure. I have taken 3 screeshots. One si for android phones like Samsung S9, another one for iPhone 6/7/8 (which is the only one that works correctly for me, like the screenshot you showed) and another screenshot for iPhone X/XS etc.

Screen Shot 2020-09-16 at 16 52 29
Screen Shot 2020-09-16 at 16 52 44
Screen Shot 2020-09-16 at 16 52 50

stanleypliu:
I don't know why I couldn't see this in Dev Tools, but increased it anyway just to be safe as I'm aware things look different depending on hardware etc.

Levia:
Works great now! Merging

PR #527, opened 2020-09-16
Ticket 208

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/208

Added social sharing links to news pages as they were previously missing, and fixed the sharing links so they contain the URL of the current page to be shared.

### DIFF
--- a/app/views/partials/social/_social-share.html.erb
+++ b/app/views/partials/social/_social-share.html.erb
@@ -1,6 +1,6 @@
 <div class="social--media">
   <% t('social.share').keys.map(&:to_s).each do |social_medium| %>
-    <%= link_to '', t("social.share.#{social_medium}.url", url: social_medium ),
+    <%= link_to '', t("social.share.#{social_medium}.url", url: url_encode(request.base_url) ),
       title: t("social.share.#{social_medium}.title"),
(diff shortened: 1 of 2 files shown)


app/views/partials/social/_social-share.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="social--media">
   <% t('social.share').keys.map(&:to_s).each do |social_medium| %>
-    <%= link_to '', t("social.share.#{social_medium}.url", url: social_medium ),
+    <%= link_to '', t("social.share.#{social_medium}.url", url: url_encode(request.base_url) ),
```
Levia, line 3: @stanleypliu `request.base_url` won't cover the intent of sharing a specific resource page but will always refer to the home page regardless. I guess this needs changing

stanleypliu: Apologies, at the time of making this PR I wasn't sure as to whether we wanted to share the base website or the current page, and then I forgot to amend this. Good catch

Levia:
Seems to work well now. Merging with the caveat that needs to be properly tested on staging

PR #528, opened 2020-09-16
Ticket 194

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/194 

Seems to work locally, the metatag image, title, description and URL all change depending on what page is being viewed


app/helpers/opengraph_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+    image.blank? ? fallback_image : path_to_image
+  end
+
+  def og_title
+    social_title = cms_fragment_content(:social_title, @cms_page)
+    title = @cms_page.label
+    fallback_title = title.blank? ? t('meta.site.title') : title
+    social_title.blank? ? fallback_title : social_title
+  end
+end
```
Levia, line 25: @stanleypliu I wonder if this is almost the same thing that the ComfyOpengraph class in `lib/modules` was trying to achieve?
If not, perhaps we could still amend or make use of that already existing class instead of creating a separate helper?

Levia: More specifically, I would also suggest to check for blanks early within the methods as otherwise the code will throw an exception for all non CMS pages when trying to fetch the label for example, as @cms_page will be nil

lib/modules/comfy_opengraph.rb
```diff
@@ -29,21 +31,62 @@ def parse(opengraph: nil, page: nil)
...
+  def og_description
+    social_desc = cms_fragment_content(:social_description, @page)
+    summary = cms_fragment_content(:summary, @page)
+    fallback_summary = summary.blank? ? I18n.t('meta.site.description') : summary
+    social_desc.blank? ? fallback_summary : social_desc
+  end
+
+  def og_image
+    image = cms_fragment_content(:image, @page).try(:attachments)&.first
+    path_to_image = URI.join(root_url, url_for(image))
```
Levia, line 81: If `image` is nil this line would probably fail or return an undesired behaviour

Levia:
Works well! Merging!

PR #529, opened 2020-09-16
Ticket 206

Levia:
## Description

* Add region and countries thumbnails in areas search
* Add delimiter to numbers in areas search

## Notes

Not sure why Stanley's commit still appears even though I'm sure that has been pushed to refresh 🤷‍♂️ 
[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/206)

### DIFF
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -4,6 +4,8 @@ class AssetsController < ApplicationController
       image = AssetGenerator.protected_area_tile(protected_area)
     elsif params[:type] == "country"
       image = AssetGenerator.country_tile(country)
+    elsif params[:type] == "region"
+      image = AssetGenerator.region_tile(region)
     else
       raise_404
     end
@@ -21,5 +23,9 @@ class AssetsController < ApplicationController
 
   def country
     @country ||= Country.where(iso: params[:id]).first
+  end
+
+  def region
+    @region ||= Region.where(iso: params[:id]).first
   end
 end
--- a/app/controllers/assets_controller.rb
+++ b/app/controllers/assets_controller.rb
@@ -1,14 +1,12 @@
 class AssetsController < ApplicationController
+  TYPES = %w(protected_area country region).freeze
+
   def tiles
-    if params[:type] == "protected_area"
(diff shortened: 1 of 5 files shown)


app/controllers/assets_controller.rb
```diff
@@ -4,6 +4,8 @@ def tiles
       image = AssetGenerator.protected_area_tile(protected_area)
     elsif params[:type] == "country"
       image = AssetGenerator.country_tile(country)
+    elsif params[:type] == "region"
+      image = AssetGenerator.region_tile(region)
```
stanleypliu, line 8: Would this be a good situation to use a case..when statement instead of if..else? From a readability perspective, I don't think it makes any difference performance-wise

stanleypliu:
Region and country thumbnails show, which is nice to see. For the region thumbnails though, they appear quite zoomed-out - is there any option in the Mapbox API to adjust the zoom level?

app/controllers/assets_controller.rb
Levia: Thanks. I've refactored a bit differently now.
I didn't use a case statement in the end as I found I could make use of the fact that method names and area types name are similar.

stanleypliu: Nicely done 🔥

app/helpers/application_helper.rb
```diff
@@ -90,6 +91,8 @@ def country_cover(country, with_tag: true)
   end
 
   def region_cover(region, with_tag: true)
+    version = Rails.application.secrets.mapbox[:version]
```
stanleypliu, line 94: I was looking for an equivalent to `before_action` for helpers that would allow you to set the version for each of the methods that require it, or at least a private method, but this isn't a controller. It's not actually that necessary and it likely doesn't exist but just for my own understanding. 

Also, saw a TODO about passing CSS classes to params? So you want to be able to extract the CSS class from the params, like `params[:css_class]` for example?

Levia: Yeah given this is not a controller it would require the class to be rewritten a bit perhaps.

With regards the comment, I think that's a very old one (before my time) so I think I'll remove it.

Thanks!

PR #530, opened 2020-09-16
Exclude gl candidates

stanleypliu:
Candidate GL sites are now excluded from the GL coverage growth chart, so only GL and relisted sites are present, also there is a small fix for the 'top 10 per region' chart because a line in the method used to build the chart was erroring as there were more than 10 countries with GL sites in a particular region. That chart actually has fixes for it in another branch/PR, I didn't want to clutter this one.

### DIFF
--- a/app/models/protected_area.rb
+++ b/app/models/protected_area.rb
@@ -43,6 +43,11 @@ class ProtectedArea < ApplicationRecord
     where.not(green_list_status_id: nil)
   }
 
+  scope :non_candidate_green_list_areas, -> {
+    includes(:green_list_status)
+    .where.not(green_list_statuses: {status: 'Candidate'}, green_list_status_id: nil)
+  }
+
   scope :most_protected_marine_areas, -> (limit) {
     where("gis_marine_area IS NOT NULL").
     order(gis_marine_area: :desc).limit(limit)
@@ -87,7 +92,7 @@ class ProtectedArea < ApplicationRecord
     # Takes an optional start year from which to start counting
     coverage_growth_hash = {}
 
-    areas = ProtectedArea.green_list_areas.where.not(legal_status_updated_at: nil)
+    areas = ProtectedArea.non_candidate_green_list_areas.where.not(legal_status_updated_at: nil)
     
     if start_year
       date_from_year = "#{start_year}-01-01 00:00:00".to_time 
@@ -307,7 +312,7 @@ class ProtectedArea < ApplicationRecord
     )
   end
 
-  OECM_ATTRS_ERROR = "'conservation_objectives' and 'supplementary_info' can only be assigned if this is an OECM area".freeze
+  # OECM_ATTRS_ERROR = "'conservation_objectives' and 'supplementary_info' can only be assigned if this is an OECM area".freeze
   def oecm_attributes
     return if is_oecm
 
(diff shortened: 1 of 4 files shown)


app/models/protected_area.rb
```diff
@@ -100,6 +105,7 @@ def self.greenlist_coverage_growth(start_year = nil)
       # year = date.to_date.year
       coverage_growth_hash[date] ||= []
       
+      # TODO - ask Ed how exactly the total GL area in the CSV is calculated...
```
Levia, line 108: @stanleypliu Does this mean we would still need to make some changes to the calculations before merging this potentially?

stanleypliu: So this comment was initially me wondering whether there was an incorrect calculation in the method somewhere because of the difference between the black box data and the coverage chart data. Subsequently having looked into the calculations further, I've found something strange - taking a date at random (I chose 2009), and then assigning `area_sum` to `areas.where("legal_status_updated_at <= ?", date).map(&:name)` I found the following sites: 
```
["Seoraksan", "Odaesan", "Gorgona", "Parco Nazionale Del Gran Paradiso", "Galeras", "Tatama", "Zona marina del Archipiélago de Espíritu Santo", "Chingaza", "Cerbère - Banyuls", "Marais D'Episy", "Pyrénées", "Guadeloupe", "Amarakaeri", "Ajloun Forest Reserve", "Lewa Wildlife Conservancy", "Ol Kinyei", "Cordillera Azul", "Wadi Al-Hitan (Whale Valley)", "Bois Du Loc'H", "Terres Australes Françaises", "Orlu", "Ras Mohammed", "Shouf", "Azraq Oasis", "Vosges Du Nord", "Malpelo", "Isla San Pedro Mártir", "Champ Du Feu", "Hochfeld", "Arakwal", "Montague Island", "Cape Byron", "Iroise", "Ol Pejeta", "Ecrins"]
```
There are no duplicates in that array, but when I then saved that list to `array` and ran `ProtectedArea.where(name: array).count` I got 43, but the length of `array` is 35. I'm suspicious that duplicates are being introduced during summation.

stanleypliu: After I looked again and mapped the array to WDPA ID instead of name, I found that it matched with the length of the original array (the array that's being reduced to the total gis area) and I don't think it is including duplicates in the summation. It's just something I was afraid of

app/models/protected_area.rb
```diff
@@ -90,23 +90,23 @@ def is_green_list
...
-
     areas = ProtectedArea.non_candidate_green_list_areas.where.not(legal_status_updated_at: nil)
     
+    sorted_dates = areas.pluck(:legal_status_updated_at).sort { |a,b| b <=> a }.uniq
+
     if start_year
       date_from_year = "#{start_year}-01-01 00:00:00".to_time 
-      areas = areas.where("legal_status_updated_at >= ? ", date_from_year)
+      sorted_dates.filter! { |date| date >= date_from_year }
     end
```
Levia: I would do all of this with just one SQL/ActiveRecord query for better performance and readability.
Both filtering (including non candidate sites) and sorting can be done within the same query.

Ideally, the entire process of cumulative sum could be achieved with a single query;  that's a little more complicated but feel free to give it a go if you like :)

Levia:
@stanleypliu I've made some amendments to the query if you would like to have a look. Feel free to dm me any question  you might have.

@stacytalbot Also, I've noticed that the growth chart's legend it sasys, "Number of Protected Areas" while the chart represents the area growth. What would be the most appropriate legend for this?

PR #532, opened 2020-09-18
Redirect to search page if PA found by slug

Levia:
## Description

Redirect to search page if PA is found by slug or render show page as usual if found by WDPA ID.
Raise 404 if PA is not found at all.

## Notes
[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/217)

### DIFF
--- a/app/controllers/protected_areas_controller.rb
+++ b/app/controllers/protected_areas_controller.rb
@@ -8,9 +8,12 @@ class ProtectedAreasController < ApplicationController
     
     @download_options = helpers.download_options(['csv', 'shp', 'gdb', 'pdf'], 'protected_area', id)
 
-    @protected_area = ProtectedArea.
-      where("slug = ? OR wdpa_id = ?", id, id.to_i).
-      first
+    # If found by slug, redirect to search page
+    # This is to overcome possible issues with PAs with same name/slug and different WDPA ID
+    pa = ProtectedArea.find_by(slug: id)
+    redirect_to search_areas_path(search_term: pa.name) and return if pa
+
+    @protected_area = ProtectedArea.find_by(wdpa_id: id.to_i)
 
     @protected_area or raise_404
 


app/controllers/protected_areas_controller.rb
```diff
@@ -8,9 +8,12 @@ def show
     
     @download_options = helpers.download_options(['csv', 'shp', 'gdb', 'pdf'], 'protected_area', id)
 
-    @protected_area = ProtectedArea.
-      where("slug = ? OR wdpa_id = ?", id, id.to_i).
-      first
+    # If found by slug, redirect to search page
+    # This is to overcome possible issues with PAs with same name/slug and different WDPA ID
+    pa = ProtectedArea.find_by(slug: id)
+    redirect_to search_areas_path(search_term: pa.name) and return if pa
```
stanleypliu, line 14: Is this just for redirects for the named sites on the original site? If so, it should be fine, but I've discovered odd behaviour, for example, if you have `/national-park` as the URL, it will return results in Russian, and the query string will also change as well.

Levia: That's an interesting scenario!
So it returns things in russian because the name of the site (`pa.name`) is actually in russian, while the slug is `national-park`. I'm using the name of the site to populate the search term if a site with the given slug is found.

I didn't envision this, but perhaps we could still use the slug (so changing `pa.name` with `id`) to populate the search term even though it would look a bit strange. Elasticsearch seems to map `-` just fine so search works anyway.

stanleypliu: I don't think it should be an issue, anything else would just get caught by the 404 page (which I'm creating). I believe it's only because users of the current website may have bookmarked these sites, they wouldn't have bookmarked www.protectedplanet.net/national-park because that isn't actually used by the site routing, only www.protectedplanet.net/167121 because that was the URL that initially took them to that page. It's only the explicitly named Green List sites that are covered (the links to all the sites you can see on this page use the slug of the site rather than the WDPA ID: https://www.protectedplanet.net/search?is_green_list=true)

Levia: https://www.protectedplanet.net/national-park could have been bookmarked because it's a URL the site redirects to when using the search funcionality.
We could leave it as it is or use the slug instead. It won't be perfect both ways anyway :)

PR #533, opened 2020-09-18
Import ready fixes

Levia:
## Description

This contain necessary fixes for the monthly import so to include OECM data correctly.
Moreover, a fix for the staging seeds rake task has been added, which I would also like to test properly again.

## Notes
This includes the changes that have been added to the refresh-copy branch for testing. So the importer with the OECM data has been tested and worked correctly


stanleypliu:
Do you need me to look at this even if it's DNM?

Levia:
Feel free to have a look if you like. If there will be additions to this I will let you know. Thanks!

stanleypliu:
Trying to run the import locally, I got an error when importing the seeds: 

```
2020-09-22T08:41:28.438Z 57235 TID-ovs51uxsv ImportWorkers::S3PollingWorker JID-34cf3932fb5001c890d64b4f INFO: fail: 37.991 sec
2020-09-22T08:41:28.438Z 57235 TID-ovs51uxsv WARN: {"context":"Job raised exception","job":{"class":"ImportWorkers::S3PollingWorker","args":[],"retry":false,"queue":"import","backtrace":true,"jid":"34cf3932fb5001c890d64b4f","created_at":1600764050.447614,"enqueued_at":1600764050.447656},"jobstr":"{\"class\":\"ImportWorkers::S3PollingWorker\",\"args\":[],\"retry\":false,\"queue\":\"import\",\"backtrace\":true,\"jid\":\"34cf3932fb5001c890d64b4f\",\"created_at\":1600764050.447614,\"enqueued_at\":1600764050.447656}"}
2020-09-22T08:41:28.439Z 57235 TID-ovs51uxsv WARN: NoMethodError: private method `select' called for nil:NilClass
2020-09-22T08:41:28.439Z 57235 TID-ovs51uxsv WARN: /Users/stanleyl/ProtectedPlanet/lib/modules/stats/country_statistics_api.rb:50:in `block in import'
/Users/stanleyl/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/activerecord-5.2.0/lib/active_record/relation/delegation.rb:41:in `each'
/Users/stanleyl/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/activerecord-5.2.0/lib/active_record/relation/delegation.rb:41:in `each'
```

Specifically it was failing on `kba_data`, so it wasn't being fetched properly.

Levia:
@stanleypliu I think this is ready to be looked and merged if you are happy with it.
If there will be further changes those could be done separately

stanleypliu:
- Revamped indexing works fine. 
- I've already looked on the staging copy and verified that the OECM attributes were present on the ProtectedArea model, so well done. Only thing I'd suggest would be to have a condition in place to handle falsy values in the importer for these OECM attributes like the empty string I saw earlier. 

Can't test the SSH change

Levia:
Thanks @stanleypliu . Could you please clarify the following points?

1. About the OECM attributes, could you please clarify where the empty string is exactly within the database and what should have been the actual value?
2. With regards the indexing, I have experience elasticsearch doesn't always index everything correctly in development. So I wonder if you have tried again with rerunning the indexing within this same branch? I can't see how the changes I made could change the indexing behaviour so much, as I've only changes where the actual process runs

stanleypliu:
On staging-copy: 
```
2.6.3 :004 > ProtectedArea.pluck(:supplementary_info).uniq
 => ["Not Applicable", "Not Reported", ""] 
2.6.3 :002 > ProtectedArea.pluck(:conservation_objectives).uniq
 => ["Not Applicable", "Not Reported", ""] 
```

Levia:
Ah I see. Empty values for those two fields are only related to non-OECM sites, so this is not really an issue I would say because those fields won't be shown on the site page. Those attributes are only shown if the site is an OECM site.

Although it's true that data seems inconsistent for non-OECM sites as you could encounter both empty and 'Not Applicable' value for those two attributes. This is something that would probably need to be addressed by the data provider, so I would raise this with them to double check.

In the meantime, if you agree, considering what I stated above I wouldn't really mark this as an issue for now as that data won't be visible on the site anyway.

stanleypliu:
Fair enough. I'm going to merge now

PR #534, opened 2020-09-18
Fixing top 10 chart

stanleypliu:
Whilst working on the coverage chart PR, I noticed there was an issue with my local top 10 chart on the Green List page, namely too many columns or not enough for certain regions depending on how many countries in that particular region had Green List sites. This PR aims to:

1) Fix the issue - also improving how it handles no countries for a particular region.

2) Refactor code as my original methods were quite hard to read and weren't written very cleanly. Also, removed redundant code which remained from the original repository and have fixed the view links for Green List sites (on the individual site pages) on another branch, as I was under the mistaken impression that there was a separate show page for Green List sites, but now I know that is the responsibility of the ProtectedAreas controller.


stanleypliu:
I have just noticed that on the GL page we have a partial which shows 3 example GL sites and the total number of GL sites which corresponds to the number of non-candidate sites (38), but on the search pages for GL sites all the GL sites, candidate and non-candidate are included which amount to 103. Should we also limit the search results as well, or conversely should we update the number of GL sites in the view link text?

app/controllers/green_list_controller.rb
```diff
@@ -89,35 +70,26 @@ def most_protected_areas
...
+      if top_countries[:countries].count < CHART_SIZE # Always return an array 10 items
+        if top_countries[:countries].empty?
+          # Create an array of empty hashes
+          top_countries[:countries] = Array.new(CHART_SIZE, {})
+        else
+          top_countries[:countries] = top_countries[:countries].in_groups_of(CHART_SIZE, {}).flatten
+        end
       end
-
+
```
Levia, line 81: I would move this logic (including the CHART_SIZE=10 constant, which could also benefit of having `.freeze` added at the end), into the region presenter.

app/presenters/region_presenter.rb
```diff
@@ -58,18 +58,26 @@ def total_polygons_percentage
...
+      polygons: begin
+                  ((polygons_count / total.to_f) * 100).round
+                rescue StandardError
+                  0
+                end,
+      points: begin
+                ((points_count / total.to_f) * 100).round
+              rescue StandardError
+                0
+              end
```
Levia, line 80: Any particular reason for adding the entire begin-rescue blocks here?
I personally find this less readable while achieving the same thing.

Potentially this could be refactored to not use rescue at all and check the value of the variables before doing calculations

app/presenters/region_presenter.rb
```diff
@@ -158,15 +149,25 @@ def top_gl_coverage_countries
...
+    countries.map do |country|
+      total_area = country.country_statistic.total_area
+      total_gl_area = country.total_gl_coverage
+      percentage_of_total_area = ((total_gl_area / total_area).to_f * 100).round(1)
+      
+      { country: country, total_area: total_gl_area, percentage: percentage_of_total_area }
+    end.sort! do |a, b|
+      # If rounded %s happen to be the same, then sort by area
+      b[:percentage] == a[:percentage] ? b[:total_area] <=> a[:total_area] : b[:percentage] <=> a[:percentage]
+    end.take(10)
```
Levia, line 162: I think `take` doesn't take into account the order of the elements in the array. I would use `first(10)` to be sure.
The overall method could perhaps benefit of a SQL version refactoring but that's ok for now

Levia (changes requested):

app/presenters/region_presenter.rb
stanleypliu: Did a test: 
```
irb(main):001:0> array = [1, 2, 3, 4, 5]
=> [1, 2, 3, 4, 5]
irb(main):002:0> array.take(2)
=> [1, 2]
irb(main):003:0> array.first(2)
=> [1, 2]
```
Looks to be the same to me?

Levia: Yeah so it should be fine with arrays; order is not preserved when using take with AR models apparently. So that's ok :)

Levia:
Works well. Merging!

PR #537, opened 2020-09-23
Ticket 221

Levia:
## Description

Fix issues related to CMS custom categories:

* Ensure categories related tables have been created before doing anything in the `after_save` method.
* Ensure categories data exist as soon as tables are created so that CMS seeds import works fine too. (See [related DB PR](https://github.com/unepwcmc/protectedplanet-db/pull/44) for this)

## Notes
[Related DB PR](https://github.com/unepwcmc/protectedplanet-db/pull/44)
[Codebase Ticket](https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/221)

### DIFF
--- a/lib/tasks/cms_categories.rake
+++ b/lib/tasks/cms_categories.rake
@@ -12,4 +12,11 @@ namespace :cms_categories do
       end
     end
   end
+
+  task destroy: :environment do
+    Comfy::Cms::LayoutsCategory.destroy_all
+    Comfy::Cms::PagesCategory.detroy_all
+    Comfy::Cms::PageCategory.detroy_all
+    Comfy::Cms::LayoutCategory.destroy_all
+  end
 end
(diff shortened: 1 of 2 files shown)


lib/tasks/cms_categories.rake
```diff
@@ -12,4 +12,11 @@ namespace :cms_categories do
       end
     end
   end
+
+  task destroy: :environment do
+    Comfy::Cms::LayoutsCategory.destroy_all
+    Comfy::Cms::PagesCategory.detroy_all
+    Comfy::Cms::PageCategory.detroy_all
```
stanleypliu, line 19: There's a couple of typos here - also was wondering if we could set this Rake task to run prior to seeds import perhaps in the Comfy monkey patch? I end up forgetting that this task exists which is probably what I should have done before trying to migrate as it's purely dependent on the contents of the YML file

Levia: Thanks! Fixed the typos.

For clarity, I'm not calling the destroy task in any migration, but just the import one which hopefully should fix the issues you pointend out

PR #539, opened 2020-09-23
404 page

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/216

404 page created with news articles, resource cards and thematic carousel all on one page, plus a redirect back to the homepage. 

I'd like to wait for Charlotte's feedback first though before review


stanleypliu:
@Levia can be reviewed now

app/controllers/application_controller.rb
```diff
@@ -16,6 +16,14 @@ class PageNotFound < StandardError; end;
   before_action :check_for_pdf
   after_action :store_location
 
+  NO_REDIRECT = ["/users/sign_in",
+    "/users/sign_up",
+    "/users/password/new",
+    "/users/password/edit",
+    "/users/confirmation",
+    "/users/sign_out"
+  ].freeze
```
Levia, line 25: I wonder if we should get rid of these as I don't think users are still able to log in.
This is something that precedes me as well but looking at the current website those pages look a bit broken :)

stanleypliu: Didn't know it was a holdover, and yes I know :)

Levia:
Looks good! Merging

PR #540, opened 2020-09-23
Ticket 207

Levia:
## Description

Given we have now got more global stats that need to be imported from a CSV, I've done the following:

* Created a new database table called `global_statistics` which will contain only one row with all the stats coming from the `global_stats.csv`. This file is originally given to us with the name like `PP_National_Monthly_Stats_Oct20.csv`, so that needs to be renamed and checked before beign pushed in the repo (more or less like we would do with the marine_statistics.csv and green_list_data.csv)
* I've removed the MarineStatsImporter altogether and refactored the GreenListImporter so that the data is imported in bulk using only the new GlobalStatsImporter I've created.
* I've refactored the bits of the monthly release importer to use the new GlobalStatistic importer and removed the MarineStatsImporter
* I've removed references to those old importers and code from the tests, although those are not guaranteed to work as we know we need to work on tests properly at some point in the future.
* I've updated the percentages and stats used across the home page, marine page and GreenList page to use the stats we get from the global_stats.csv as specified.

## Notes
[Related PP DB PR](https://github.com/unepwcmc/protectedplanet-db/pull/45)
[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/207)

Monthly release importer would need to be tested on staging to ensure everything still works as intended.
To update your local db with the data:
```
> Wdpa::GlobalStatsImporter.import
```

As you will notice, in this way we got rid of redis with regards saving stats there.
I decided to go this route as I find the stats simpler to manage using Rails and also we could further improve it in the future to have some more immediate feedback from the database if the CSV we are given doesn't follow the value types.
Moreover, if these stats will ever be included in the PP-API, it's easier to fetch those from the shared db rather then from redis.

Any questions or concerns welcome!


stanleypliu:
I agree that it is a better idea to have a single store of statistics, I decided to go with Redis as a store for the Green List data because I wasn't sure where it could go, short of creating a singleton class like you've done here. Obviously it has downsides in that it's not as easy to access said stats, nor work with them if needed. However, good job consolidating the statistics into one class. 

What about the methods defined in the Stats::Global module itself? Are there any which would now be redundant or need to be re-evaluated in light of your new class?

Levia:
Thanks!

Yeah, that's a good point.
I've deliberately left those untouched because calculations should be correct even though based on the ProtectedArea and CountryStatistic records data. I would perhaps create a low priority ticket to address those in the future while we focus on the launch for now. What do you think?

stanleypliu:
That's fine by me - I feel that we have to take some time afterwards to properly compare our database data with the CSV (CLS' own data) to find out what is causing discrepancies.

app/models/global_statistic.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  GREEN_LIST_STATS = %w(
+    green_list_perc
+    green_list_area
+    green_list_count
+  ).freeze
+  def self.green_list_stats
+    instance.slice(*GREEN_LIST_STATS)
+  end
+
```
stanleypliu, line 36: Do you think there should be an accessor style method here as well for land stats, with corresponding array of strings?

Levia: Yeah there could be, although I don't see how it would be used at the moment?
Most of the stats (if not all) are accessed via the marine and greenlist ones and directly from a specific field in the table

stanleypliu: That's fine then, merging

PR #543, opened 2020-09-24
Refresh with master

Levia:
## Description

This is a candidate branch to become the future master soon.
It has the work of the refresh branch merged in with master and with resolved conflicts.

Further work merged into `refresh` would also need to be merged here, but if we realise this branch works as intended we wcould make this the new `refresh` or perform the same changes into the actual `refresh` branch. Hope it's not too confusing!

I've added this also so that we could test this on staging to ensure everything still works as intended even after fixing the conflicts with master.

🔴 Do not merge yet please!


stanleypliu:
Just a note that this branch is currently deployed to staging-copy

PR #544, opened 2020-09-24
Map zoom

WKocur:
I have some of Ferdi's commits that I've pulled over from refresh... Not too sure what the deal is with refresh and refresh copy. Let me know if you want me to cherry-pick my commits and pop them on another branch. To be honest, it's just two line changes so very easy!

default-options.js and mixin-add-layers.js are the two files I changed. 

One sets the initial zoom of the map (and removes the default bounding box as this will overwrite the zoom)
The other sets the tile size to match the arcgis tilesize to make the layers less blurry.


stanleypliu:
Not sure either, but your changes look good. Merging

PR #545, opened 2020-09-25
adjust initial center location for oecm map

WKocur:
This should centre the map on roughy mexico on the oecm page. Other maps should be unaffected. 

Worth checking the oecm, wdpa, marine and home pages to check there are no issues with undefined local variables etc.

### DIFF
--- a/app/controllers/oecm_controller.rb
+++ b/app/controllers/oecm_controller.rb
@@ -22,6 +22,9 @@ class OecmController < ApplicationController
         { url: OECM_FEATURE_SERVER_LAYER_URL, isPoint: false }
       ]
     }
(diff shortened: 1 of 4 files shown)


stanleypliu:
Map isn't centred on Mexico unfortunately (I'm on the OECM page):

Screenshot 2020-09-25 at 14 25 14

stanleypliu:
Other maps are unaffected but I had to explicitly insert parentheses in the `local_assigns.has_key?` method call to stop it from erroring.

stanleypliu:
Now map appears as it should on the OECM page - I believe webpack-dev-server did not recompile to reflect your changes. Merging

PR #546, opened 2020-09-27
Fix/og image urls

Levia:
## Description

Issues with host not being set when fetching images url came up again.
Also, `AssetUrlHelper` module needs to be included so to use the `image_path` method; not sure why this seemed to have worked without it before.

These changes have already been temporarily applied to the copy staging instance without deploying just to ensure it works.
Withouth these changes, pages like the marine page and the GL pages were throwing an exception.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -120,8 +120,8 @@ class ApplicationController < ActionController::Base
   end
 
   def set_host_for_local_storage
-    Rails.application.routes.default_url_options[:host] = request.base_url if Rails.application.config.active_storage.service == :local
+    Rails.application.routes.default_url_options[:host] = request.base_url
     # TODO Check why this is not set automatically
-    ActiveStorage::Current.host = request.base_url if Rails.application.config.active_storage.service == :local
+    # ActiveStorage::Current.host = request.base_url if Rails.application.config.active_storage.service == :local
   end
 end
--- a/lib/modules/comfy_opengraph.rb
+++ b/lib/modules/comfy_opengraph.rb
@@ -3,6 +3,7 @@
 # Helper class to set Opengraph meta-tags for use with Comfy CMS @cms_page's
 class ComfyOpengraph
   include Rails.application.routes.url_helpers
+  include ActionView::Helpers::AssetUrlHelper
   include Comfy::CmsHelper
 
   attr_accessor :mappings


stanleypliu:
About your comment concerning `image_path`, I've never experienced that before myself, perhaps the issue was present but we didn't notice it. 

What do you mean when you say those pages were not throwing an exception? I think they should continue to load even though the Opengraph image tag isn't being set

Levia:
Sorry, I meant to say that the were indee throwing an exception (fixed the description above now).
So basically, without these changes, the system was complaining about the following, in order:

1. host parameter not set
2. image_path is undefined

app/controllers/application_controller.rb
```diff
@@ -120,8 +120,8 @@ def check_for_pdf
   end
 
   def set_host_for_local_storage
-    Rails.application.routes.default_url_options[:host] = request.base_url if Rails.application.config.active_storage.service == :local
+    Rails.application.routes.default_url_options[:host] = request.base_url
```
stanleypliu, line 123: Could this go in `environment.rb` seeing as it no longer depends on the environment? It technically is a config setting.

stanleypliu:
I've just tested myself locally and on staging-copy, the differences with and without your changes which I reverted. I can't see anything different, the page still loads without your changes, but those pages on staging-copy don't actually have hero images, hence the OG image tags will just default to the screenshot of the website. I remember this problem cropping up before, and I wonder if there aren't settings we can tweak in `storage.yml` as well for ActiveStorage.

app/controllers/application_controller.rb
Levia: yeah I guess we could. I haven't tested it myself but feel free to give this a go. 
Also probably a matter of making sure the host url to use is pulling the correct one from the env vars

Levia:
I don't know if that could possibly related to caching. So worth testing after clearing the cache as well.
Although I don't think this issue is related to the ActiveStorage; it seems more related to a combination of missing CMS files and standard Rails configuration

stanleypliu:
Yeah I've just seen the issue on those pages - seems independent of me clearing the cache though as it seems it's cached by the server rather than the Rails application. Also, I've just realised that it's better to override the `default_url_options` in ApplicationController rather than the individual configuration files, since there is no `routes` config option and I would need to insert that line into each environment configuration file. 

However, I did notice that it's erroring on the `social.png` file which isn't a CMS file?

stanleypliu:
Anyway, I'll merge as it works fine as it is

PR #547, opened 2020-09-29
Oct release

stanleypliu:
This are the files I've been given for the Oct release, and also I've tweaked the new `GlobalStatsImporter` to work for future releases.


lib/modules/wdpa/global_stats_importer.rb
```diff
@@ -4,14 +4,16 @@ module Wdpa::GlobalStatsImporter
...
     attrs = {singleton_guard: 0}
     CSV.foreach(GLOBAL_STATS_CSV, headers: true) do |row|
       field = row['type']
       value = parse_value(row['value'])
       attrs.merge!("#{field}": value)
     end
-
-    GlobalStatistic.create(attrs)
+    
+    GlobalStatistic.first_or_create(attrs)
```
Levia, line 16: If you go for the `destroy_all` option above, you could just leave `create` as `first` will always be nil anyway

stanleypliu: Left that in by mistake - how about this new method? Should now update in place if it differs

lib/modules/wdpa/global_stats_importer.rb
```diff
@@ -11,7 +13,13 @@ def self.import
...
     end
 
-    GlobalStatistic.create(attrs)
+    stats = GlobalStatistic.first_or_create(attrs)
+    init_stats = GlobalStatistic.new(attrs)
+
+    # Destroy GlobalStatistic if it differs and recreate
+    if stats.attributes.except!(*EXCLUDED_ATTRS) != init_stats.attributes.except!(*EXCLUDED_ATTRS) 
+      stats.update(attrs)
+    end
```
Levia, line 22: This is unnecessarily overcomplicated.
The previous approach was ok, it was just a matter of putting back `GlobalStatistic.create()` rather then using `first_or_create`.

Using `first_or_create` isn't probably the best because you would need to account for when the object already exists, so just passing the attributes to that method won't work in that case. Hence why then you had to adapt the algorithm.

Another possible approach would be something like:
```
stats = GlobalStatistic.find_or_initialize_by(singleton_guard: 0) #singleton_guard is mandatory
stats.update_attributes(attrs) #works with the already existing object or the one just initialized with singleton_guard = 0
```

You could test it out in the rails console to have a better understanding of how that works

stanleypliu: I didn't want to destroy the GlobalStatistic instance unnecessarily if I could just update it (being a singleton class after all), so I opted for the above. Given what you've said, I am looking at using `first_or_initialize` now which seems to be analogous to `find_or_initialize_by` - either way you're right, this does save unnecessary code

PR #548, opened 2020-09-29
added mpa map to marine protected areas page

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/227

Added the MPA map to the dropdown list of downloads on the Marine Protected Areas page which will open in a new tab

### DIFF
--- a/app/helpers/downloads_helper.rb
+++ b/app/helpers/downloads_helper.rb
@@ -15,6 +15,11 @@ module DownloadsHelper
       title: 'File Geodatabase',
       commercialAvailable: true
     },
+    mpa_map: {
+      isDownload: false,
+      title: 'MPA Map',
+      url: URI.join(Rails.application.routes.url_helpers.root_url, 'MPA_Map.pdf')
+    },
     esri_wdpa: {
       isDownload: false,
       title: 'ESRI Web Service',
@@ -44,7 +49,7 @@ module DownloadsHelper
   end
 
   def download_params(format, domain, token)
-    return {} if %w(esri_wdpa esri_oecm).include?(format)
+    return {} if %w(esri_wdpa esri_oecm mpa_map).include?(format)
     _domain = format == 'pdf' ? 'pdf' : domain
     {
       params: {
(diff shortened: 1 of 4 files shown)


app/helpers/downloads_helper.rb
```diff
@@ -15,6 +15,11 @@ module DownloadsHelper
       title: 'File Geodatabase',
       commercialAvailable: true
     },
+    mpa_map: {
+      isDownload: false,
+      title: 'MPA Map',
+      url: URI.join(Rails.application.routes.url_helpers.root_url, 'MPA_Map.pdf')
```
Levia, line 21: @stanleypliu Does this work for you? I have an error regarding the root_url which cannot be found.
However, I fell that's a bit unnecessary as the root url with the link tag can be achieve with just `/MPA_Map.pdf` which automatically fetches the file from the public folder.

Morover, I wonder if we should have a controller action that actually makes a download available for the user rather then opening the file up on the browser? I guess this is more a design question for @stacytalbot 
Otherwise we can just skip that for now to speed things up

stacytalbot: Good point - if it doesn't take too long then add as a dl like all the others.

stanleypliu: Have done, added a new key called `isMap` which I then also added in the `DownloadPopup` component and set the `download` HTML5 attribute to download the file named as it currently is.

app/javascript/components/download/DownloadPopup.vue
```diff
@@ -11,6 +11,14 @@
...
           v-html="option.title"
         />
+        <a
+          v-else-if="option.isMap"
+          class="popup__link"
+          :href="option.url"
+          :download="option.title"
+          :title="option.title"
+          v-html="option.title"
+        />
```
Levia, line 21: This works well!
Although I've noticed that the map link is always underlined, perhaps we should adjust the style a bit.

Screen Shot 2020-09-29 at 16 16 31

stanleypliu: @stacytalbot Noticed that anchor tags were always set to be underlined in `base.scss` - can I remove it for `popup__link`?

stacytalbot: yea

Levia:
Works well! Merging!

PR #550, opened 2020-09-29
Ticket 164

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/pp-refresh/tickets/164

Added Type filter to PAME page to allow filtering by either Marine or Terrestrial areas. Filtering carries over to CSV download 

Need to check that Marine filter incorporates coastal areas as well.


Levia:
Works well, good job!
I wouldn't worry about coastal areas as those are currently mapped as marine anyway

PR #551, opened 2020-09-30
Reinsertion of ribbon

stanleypliu:
- Ribbon with number of sites has been reinserted in the top left of the applicable thematic carousel cards. 
- Text now positioned directly below it and at the same height for all cards regardless of whether a ribbon is present or not. 
- Carousel height has also been increased to allow summary text to show if a particular card has too much text (see the ICCA one).

### DIFF
--- a/app/views/partials/sub_partials/_themes.html.erb
+++ b/app/views/partials/sub_partials/_themes.html.erb
@@ -2,6 +2,17 @@
   title: "View the #{theme[:obj][:label]} page", 
   class: "card--theme", 
   style: "background-image: url('#{cms_fragment_render(:image, theme[:obj])}')" do %>
+  <% unless theme[:pas_no] == -1 %>
+    <div class="card__ribbon">
+      <i class="card__icon <%= theme[:obj][:slug] %>">
+        <%= render '/partials/svgs/pin.svg' %>
+      </i>
+      <p class="no-margin">
+          <span class="card__number"><%= theme[:pas_no] %></span>
+          <span class="card__number-text"><%= t('global.area-types.wdpa') %></span>
+      </p>
+    </div>
+  <% end %>
 
   <h3 class="card__title"><%= theme[:obj][:label] %></h3>
   <p class="card__summary"><%= cms_fragment_content(:summary, theme[:obj]) %></p>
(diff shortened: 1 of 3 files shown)


app/views/partials/sub_partials/_themes.html.erb
```diff
@@ -2,6 +2,17 @@
   title: "View the #{theme[:obj][:label]} page", 
   class: "card--theme", 
   style: "background-image: url('#{cms_fragment_render(:image, theme[:obj])}')" do %>
+  <% unless theme[:pas_no] == -1 %>
+    <div class="card__ribbon">
+      <i class="card__icon <%= theme[:obj][:slug] %>">
+        <%= render '/partials/svgs/pin.svg' %>
+      </i>
+      <p class="no-margin">
+          <span class="card__number"><%= theme[:pas_no] %></span>
```
Levia, line 11: Some cards that would need to have a ribbon have that missing still, like wdpa and oecm.
Worth checking the pas_figure method in the presenter is not skipping some labels

stanleypliu: Cannot remember if this ever took place, but I seem to recall working on this carousel a few months ago and adding a delimiter to the numbers. Even then there were cards missing that ribbon, not sure if this was intended. I think they may have even been taken out deliberately.

stacytalbot: Yea it doesn't make sense for some to have labels i believe. Jess wanted to be able to give us a number to put in manually for the ICCA card

stacytalbot: I told her we could add a field to the CMS if time.

stacytalbot: WDPA and OECM do need one though

stanleypliu:
Decided to change the way the data was being fetched from the backend, we were using the label of the page to determine which number to assigned to which card. However, the label is more likely to change than the slug (and that was the reason why the numbers disappeared - the label of the WDPA and OECM pages were changed), so I changed the `THEMATIC_AREAS` variable to an array of the page slugs so it's less error-prone.

Levia:
Works well. Merging

PR #556, opened 2020-10-01
Ensure to pull the correct number of other sites

Levia:
## Description

Site page - Other sites not showing correctly yet. Site with WDPA ID 555547601 only shows 2 more sites (the given site is the only site in the country).


stanleypliu:
I ended up reverting my own commit which pretty much resembled your fix entirely. Thanks for fixing it!

PR #558, opened 2020-10-01
social media previews

stanleypliu:
~~Set the Twitter card to use 'summary_large_image' and also use 4096 x 4096 (the maximum size) variants in case images are too large for Twitter~~

EDIT: After setting the staging S3 bucket to public-read, setting the Content-Type of the images and resizing the images, the S3 hosted hero images still cannot be loaded by the web crawlers used by the social media bots. I've resorted to using the default social image for all pages which can be shared - which as it turns out still doesn't work on the news pages for unknown reasons (the GL page and other thematic area pages work). There are still useful changes I've made so this should still be merged in. 

Also, apologies for the messy commit history, tried to amend commits after pushing (instead of before)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -36,8 +36,7 @@ class ApplicationController < ActionController::Base
       'image:alt': t('meta.image_alt'),
       'image:height': t('meta.image_height'),
       'image:width': t('meta.image_width'),
-      'locale': 'en_GB',
-      'locale:alternate': "#{I18n.locale}_#{I18n.locale.upcase}"
+      'locale': 'en_GB'
     }
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,22 +22,28 @@ class ApplicationController < ActionController::Base
   def opengraph
     return if admin_path?
 
-    @opengraph ||= OpengraphBuilder.new({
-                                          'og': {
-                                            'site_name': t('meta.site.name'),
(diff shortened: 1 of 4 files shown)


Levia:
@stacytalbot @stanleypliu What have we decided to do with this?
I can see there a few commented bits and TODOs (also in the case statement with regards the image)

stanleypliu:
can this be merged in? it'll return previews for *most* pages with the standard `social.png` as the image
aim to return to this when there's more budget to investigate why it's not rendering the cards for news pages - no longer convinced it's just to do with the bucket as I've commented out the code to render the custom image (the hero image if present)

app/controllers/application_controller.rb
```diff
@@ -22,22 +22,31 @@ def admin_path?
...
+      'title': t('meta.site.title'),
+      'description': t('meta.site.description'),
+      'url': request.url,
+      'type': 'website',
+      'image': URI.join(root_url, helpers.image_path(t('meta.image'))),
+      'image:alt': t('meta.image_alt'),
+      'image:height': t('meta.image_height'),
+      'image:width': t('meta.image_width'),
+      'locale': 'en_GB',
+      'locale:alternate': "#{I18n.locale}_#{I18n.locale.upcase}"
```
Levia, line 40: Why are we using an hardcoded string for locale while using I18n for locale:alternate?

stanleypliu: sorry, this line was later removed

Levia:
Merging with the caveat that this needs further fixes about showing the desired preview image

PR #561, opened 2020-10-01
Make sure nav scrolls on mobile

stacytalbot:
+ changes to equity that weren't pushed before
+ update chart on the green list page to use tooltips for both bars
+ make sure nav scrolls on mobile

### DIFF
--- a/app/assets/stylesheets/components/_nav.scss
+++ b/app/assets/stylesheets/components/_nav.scss
@@ -42,12 +42,21 @@ $pane-side-padding: 42px;
 
       &__close {
         @include button-close;
-        @include responsive(display, block, none, none);
-        margin: rem-calc(20 20 0) auto;
+
+        display: block;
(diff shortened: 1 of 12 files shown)


Levia:
Does this look correct? I'm a bit confused by the mismatch with the color of the bar chart and the legend. @stacytalbot 
Screen Shot 2020-10-06 at 14 50 24

Levia:
@stacytalbot Looks better!

There's only one thing there's a bit confusing for me still.

In the GL page, shouldn't the `% of Protected Area Coverage` correspond to the percentage we have in the home page? So 9.6% instead of 9.18%. 
Baring also in mind that 9.6 is the figure that takes into account both WDPAs and OECMs, and that the figure that we have in the GL page is calculated by subtracting the GL coverage from the WDPA (only WDPAs) coverage, which doesn't sound about right to me for this chart.

PR #562, opened 2020-10-06
Add designation link to search page

Levia:
## Description

Add designation link to search page

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -148,4 +148,8 @@ module SearchHelper
       end)
     end
   end
+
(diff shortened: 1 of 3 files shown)


stanleypliu:
Well spotted! However, I think you also need to pass the country to the filters as well, otherwise it's listing all international sites of that designation.

Levia:
@stacytalbot said that it shouldn't work like that. The goal is to redirect the user to the search page listing all sites with the selected designation, as if there was a designation page so to say.

Although I'm still looking at something with regards the fact that not all designations seem to be listed in the filters

PR #564, opened 2020-10-06
Ensure shapefile is generated correctly

Levia:
## Description

Ensure all generated files are merged into main download after data is split into 3 shapefiles

PR #565, opened 2020-10-06
Use custom layer name for gdb exports

Levia:
## Description

Rename GDB layers so they won't appear like `sql_statement`.

## Notes

Needs to be tested properly when #564 is merged


Levia:
@stanleypliu This is ready to be reviewed now. Thanks!

PR #571, opened 2020-10-07
Fix site with no country

Levia:
## Description

Sites can have no countries attached apparently.
This means that the view all button should redirect to search page with all areas instead of throwing an exception.

Also, the 'view all' button was not taking into consideration the country correctly as well. This has also been fixed.

## Notes

[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/159)

### DIFF
--- a/app/controllers/protected_areas_controller.rb
+++ b/app/controllers/protected_areas_controller.rb
@@ -130,7 +130,7 @@ class ProtectedAreasController < ApplicationController
     if area.is_transboundary
       search_areas_path(filters: { special_status: ['is_transboundary'] })
     else
-      search_areas_path(filters: { location: { type: 'site', options: ["#{@countries.first.name}"] } })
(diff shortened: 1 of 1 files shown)


stanleypliu:
Works well - I didn't know sites could belong to no countries. Noticed that we might want to also change the sample sites that are present in those cards as well as they are currently showing transboundary sites because I envisioned that there would only be 1 or more countries per site. Up to you, personally I feel it depends on how many ABNJ sites there are and whether it's worth the effort to tailor the `get_other_sites` method

Levia:
@stanleypliu That should be already accounted for.
`@countries` is defined as such:

```
@countries = @protected_area.countries.without_geometry
```

which returns an empty array if there are no countries for the given PA.
This means that `@countries.length` is 0 and so the `get_other_sites` method will get the transboundary sites.
If the there are no transboundary sites, or they are not at least 3, then the algorithm will get the remainder.

PR #573, opened 2020-10-27
Add url to gl status

stanleypliu:
Linked with https://github.com/unepwcmc/protectedplanet-db/pull/47 

In preparation for the November release of the WDPA, I added a row to the ~~GreenListStatus~~ Protected Area model for the URL:
- Green List Importer has also been updated to include the URL
- Included the new GL sites CSV 
- Updated the ProtectedArea Presenter to show styled URL on GL page if a URL is present 

Migration present so you need to migrate the db to get it working

Also an update for puppeteer is present.

👎  ~~Importer doesn't work properly so I need to fix~~ Didn't realise that multiple GL areas could belong to a Green List Status, so it was working fine


app/assets/stylesheets/components/cards/card/stats/_card-stats-affiliations.scss
```diff
@@ -4,6 +4,12 @@
 &--stats-affiliations {
   @include card-stats;
 
+  .icon--circle-chevron-black {
```
stacytalbot, line 7: I would suggest creating a &__button-external class and importing something from the button stylesheet if you can - we have that chevron on some other buttons on the country page so should be a class you can use.

app/views/partials/stats/_stats-affiliations.html.erb
```diff
@@ -16,6 +16,13 @@
               <p class="card__subtitle"><%= t('stats.green-list.date') %></p>
               <span><%= link[:date] %></span>
             <% end %>
+
+            <% if link[:url] %>
+              <a class="card__subtitle--link" href="<%= link[:url] %>" >
```
stacytalbot, line 21: add title attribute and open link in a new window

PR #575, opened 2020-11-10
Add who link

Levia:
## Description

* Add World Heritage Outlook link to PA page if PA is designated as a World Heritage Site
* Increase map size in PA page only, so it matches the height of the stats section on the left (especially if there are many links)

## Notes

@stacytalbot I've added a TODO comment with regards the `button_title` text in the locales YAML. That doesn't seem to be used so it would be good to double check what we should do with that.

[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/138)

### DIFF
--- a/app/assets/stylesheets/base/_buttons.scss
+++ b/app/assets/stylesheets/base/_buttons.scss
@@ -26,11 +26,11 @@ $padding-medium: rem-calc(27);
 @mixin button-font {
   @include responsive(font-size, rem-calc(16), rem-calc(18), rem-calc(18));
 }
-@mixin button-disabled { 
-  cursor: not-allowed; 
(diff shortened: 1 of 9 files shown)


stanleypliu:
I've gone and included the button titles again for the links just in case we want to keep them. Anyway, looks good from my side

Levia:
@stacytalbot Thanks. That's done.
I've also adjusted the width of the icon to make it more apart with the "Link" button.

Not sure this needs to be applied also to the white version of the icon as I can't see it being used?

Levia:
@stacytalbot Also just noticed the link has the underline property, which does not appear in the screenshot you shared. Do you want me to also get rid of that?

PR #577, opened 2020-11-12
Active storage transfer

Levia:
## Description

Ensure to also transfer the ActiveStorage related tables so to not lose reference to attachments (e.g. background images)

⚠️ Please test this properly before merging

### DIFF
--- a/app/workers/import_workers/finaliser_worker.rb
+++ b/app/workers/import_workers/finaliser_worker.rb
@@ -26,6 +26,7 @@ class ImportWorkers::FinaliserWorker
     # we can't get from the WDPA
     CmsTransfer.transfer
     ApiTransfer.transfer
+    ActiveStorageTransfer.transfer
 
     # Clear the redis cache and delete all downloads
     # from the S3 bucket. This will trigger the generation
--- a/lib/modules/active_storage_transfer.rb
+++ b/lib/modules/active_storage_transfer.rb
@@ -1,0 +1,28 @@
+module ActiveStorageTransfer
+  def self.transfer only_dump=false
+    config = ActiveRecord::Base.connection_config
+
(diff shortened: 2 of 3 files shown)


stanleypliu:
Finally(!) managed to import a release on the new-web-copy server instance, seems to work, all images are showing, and I can see that the IDs of the ActiveStorage attachments table is preserved between `refresh_copy` and `refresh_copy_backup`, so, good job

I do get the following errors in the console:
```
ERROR:  relation "active_storage_attachments" already exists
ALTER TABLE
ERROR:  relation "active_storage_attachments_id_seq" already exists
ALTER TABLE
ALTER SEQUENCE
ERROR:  relation "active_storage_blobs" already exists
ALTER TABLE
ERROR:  relation "active_storage_blobs_id_seq" already exists
ALTER TABLE
```
during the import of the tables. I can see that the Comfy import has even more of these errors, presumably because the existing tables aren't dropped before restoring using the dumps. However, I'm guessing this isn't new to you - any edge cases where they could cause problems?

Levia:
@stanleypliu Cool!

Yeah I believe it's because those tables might be created by default at the moment of initializing the new database (although I would need to double check) and then the transfer scripts encounters those tables (so they are not dropped beforehand).

Although, unless a flag is specified in the command, I think the SQL scripts continues doing it's job. For the pg_restore command there's a `--clean` flag that drops the tables before they are overwritten, however I couldn't find a similar flag for the `psql` command.

If you are in a hurry to get this live, I would say that the most important thing is that the tables have been copied over fine, so double checking that the active storage tables in the `_backup` db are the same as the ones in the new db.

stanleypliu:
Yes the tables do indeed match up, ID for ID, record_id for record_id and so on.

PR #578, opened 2020-11-20
stats now aligned on homepage and mpa page

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/176

Statistics are now aligned on the homepage cards and the MPA page. The MPA page now no longer takes OECMs into account in the statistics, just PAs. Likewise, the number of marine protected areas on the homepage (excluding OECMs) now uses `GlobalStatistic` to populate it, as opposed to performing a query on the ProtectedArea model.

### DIFF
--- a/app/controllers/marine_controller.rb
+++ b/app/controllers/marine_controller.rb
@@ -26,9 +26,9 @@ class MarineController < ApplicationController
       RegionPresenter.new(region).marine_coverage
     end
 
-    @pas_km = @marine_statistics['total_ocean_area_oecms_pas']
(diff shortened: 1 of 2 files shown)


stanleypliu:
I spoke with Ed and he said that I'm to use the global stats csv to populate the number of terrestrial protected areas as well. Seeing as the coverage % also comes from the csv (indeed all of the other stats) we no longer rely on the database to provide us figures for the cards, so I've gone and deleted the commented out lines in the HomePresenter as well as the comment stating that the figure for the terrestrial PAs comes from the DB.

PR #583, opened 2020-12-15
Link to tabs

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/183

Set up pages with tabs on them (Thematic Area pages) so that they now add a `tab` param to the URL query string when accessed which enables specific tabs to be directly linked to.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -91,8 +91,11 @@ class ApplicationController < ActionController::Base
   def load_cms_content
     return if admin_path?
 
-    @cms_page ||= Comfy::Cms::Page.find_by_full_path(request.original_fullpath
-                  .gsub(%r{\A/#{I18n.locale}/?}, '/')[/[^?]+/])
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -91,7 +91,8 @@ def load_cms_site
   def load_cms_content
     return if admin_path?
 
-    @cms_page ||= Comfy::Cms::Page.find_by_full_path(request.original_fullpath.gsub(%r{\A/#{I18n.locale}/?}, '/'))
+    @cms_page ||= Comfy::Cms::Page.find_by_full_path(request.original_fullpath
```
Levia, line 94: I would refactor this slightly so it's more readable.
More specifically, creating a local variable with the value that goes into `find_by_full_path` and naming it properly would also help understand what that is exactly.

Levia (changes requested):
Just one minor thing related to refactoring; everything else seems to work well!

PR #588, opened 2021-01-25
fix/invoke rails:cache:clear post-deploy:finishing

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/185

This should hopefully solve all the issues regarding missing content on pages that often occurs after a deploy. Turns out that there was already an old Rake task for clearing the Rails cache, all I've done is add it as a callback to the deploy:publishing hook. 

I've checked and `Rails.cache.clear` does not clear the Redis cache, either directly through the console, checking both the Redis keys and manually through checking the downloads on the website itself before, during and after clearing the cache. I'm not 100% sure, but I think this means that the downloads are safe and shouldn't be reset.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,5 +1,5 @@
 # config valid only for current version of Capistrano
-lock '3.14.1'
+lock '3.11.0'
 
 set :repo_url, 'git@github.com:unepwcmc/ProtectedPlanet.git'
 set :application, "ProtectedPlanet"
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,5 +1,5 @@
 # config valid only for current version of Capistrano
-lock '3.11.0'
+lock '3.14.1'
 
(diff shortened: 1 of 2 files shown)


Levia (approved):
Looks good to me!
The only thing I would like to check is the lock on the 3.14.1 version of capistrano. What version of ruby were you on?

stanleypliu:
I'm on Ruby version 2.6.3, and the strange this is that capistrano 3.11.0 is installed, but it keeps saying that 3.14.1 is loaded _only_ when I just run `cap staging deploy`. This disappears when I do `bundle exec cap staging deploy/cap production deploy` - learnt something new. I'll revert the change

Levia:
Thanks!
When you use `bundle exec` it uses the version of the gems within the Gemfile in that project folder, whereas using just `cap staging deploy` runs the version of the gems installed systemwide. So it might be you have a different (more recent) version of the capistrano Gem installed systemwide and the system was using that when peforming the deploy.

Merging now

PR #590, opened 2021-01-26
Fix/Localisation of routes

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/156

Not too sure about this one - what I've done in the end is to only transport the region and country routes out of the locale scope. I've also removed unused search endpoints and attempted to introduce some organisation in terms of JSON endpoints vs actual views. 

Would appreciate guidance on whether this is OK for now (have I moved the wrong routes or not enough)?

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,6 +14,18 @@ Rails.application.routes.draw do
   get '/admin', to: redirect('/admin/sites')
   #root to: 'home#index'
 
+  ## Non-CMS routes - no localisation
+  get '/region/:iso', to: 'region#show', as: 'region'
+
+  get '/country/:iso', to: 'country#show', as: 'country'
+  get '/country/:iso/pdf', to: 'country#pdf', as: 'country_pdf'
+  get '/country/:iso/compare(/:iso_to_compare)', to: 'country#compare', as: 'compare_countries'
+  get '/country/:iso/protected_areas', to: 'country#protected_areas', as: 'country_protected_areas'
+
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -42,54 +54,46 @@
       end
     end
     
+    ## Only CMS routes are present below
+
+    # TODO - remove? 
     # resources :projects, only: [:create, :index, :update, :destroy]
```
stacytalbot, line 60: Remove resources bit

config/routes.rb
```diff
@@ -42,54 +54,46 @@
...
 
     get '/marine/download_designations', to: 'marine#download_designations'
-    # get '/green_list/:id', to: 'green_list#show', as: 'green_list'
+
+    # TODO - remove?
+    # get '/green_list/:id', to: 'green_list#show', as: 'green_list'
```
stacytalbot, line 65: Remove green list bit

config/routes.rb
```diff
@@ -14,6 +14,18 @@
   get '/admin', to: redirect('/admin/sites')
   #root to: 'home#index'
 
+  ## Non-CMS routes - no localisation
```
stacytalbot, line 17: I'm starting to think that these region/country and site pages should stay localized because we will still be translating the non WDPA parts with yml files.

stanleypliu: Have moved them back under the scope

PR #591, opened 2021-01-26
Ticket 188 downloads name

Levia:
## Description

* PAME download filename to follow YYYY-MMM version convention
* Standardise downloads filenames and features layers within GDB.
* For GDB exports, include source table within GDB file itself as in the original monthly release GDB

~~## TODO~~

~~* Single site downloads filenames should have a suffix to identify the area downloaded~~
~~* Look into splitting shapefiles into 3 parts only when strictly necessary (e.g. currently 3 shapefiles are generated even when downloading a single site)~~

[Codebase ticket](https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/188)

### DIFF
--- a/app/javascript/components/forms/DownloadCsv.vue
+++ b/app/javascript/components/forms/DownloadCsv.vue
@@ -43,6 +43,10 @@
 
         axios.post('/pame/download', data, config)
           .then((response) => {
+            // content-disposition looks something like: 'attachment; filename="the_file_name_here.csv"
+            // so splitting the string by 'filename=" will leave us with ['attachemnt;', 'the_file_name_here.csv"']
+            // we then get the last item and split it further by the the remaining '"' to ensure anything after that is gone.
+            // Finally we get the first element of that split.
             const filename = response.headers['content-disposition'].split('filename\="')[1].split('\"')[0]
 
             this.createBlob(filename, response.data)
--- a/app/javascript/components/forms/DownloadCsv.vue
+++ b/app/javascript/components/forms/DownloadCsv.vue
@@ -43,8 +43,7 @@
 
         axios.post('/pame/download', data, config)
           .then((response) => {
-            const date = new Date().toJSON().slice(0,10),
-              filename = `protectedplanet-pame-${date}.csv`
+            const filename = response.headers['content-disposition'].split('filename\="')[1].split('\"')[0]
 
             this.createBlob(filename, response.data)
 
--- a/lib/modules/download/generators/gdb.rb
+++ b/lib/modules/download/generators/gdb.rb
@@ -8,7 +8,7 @@ class Download::Generators::Gdb < Download::Generators::Base
       select: Download::Utils.download_columns(reject: [:gis_area, :gis_m_area]),
       where: %{"TYPE" = 'Point'}
     }
-  }
+  }.freeze
 
   def initialize zip_path, wdpa_ids
     @path = File.dirname(zip_path)
--- a/lib/modules/download/generators/gdb.rb
+++ b/lib/modules/download/generators/gdb.rb
@@ -38,7 +38,7 @@ class Download::Generators::Gdb < Download::Generators::Base
   private
 
   def export_component name, props
-    component_path = gdb_component(name)
+    component_path = gdb_component
     view_name = create_view query(props[:select], props[:where])
 
     return [] if ActiveRecord::Base.connection.select_value("""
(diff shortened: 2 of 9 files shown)


app/javascript/components/forms/DownloadCsv.vue
```diff
@@ -43,8 +43,7 @@
 
         axios.post('/pame/download', data, config)
           .then((response) => {
-            const date = new Date().toJSON().slice(0,10),
-              filename = `protectedplanet-pame-${date}.csv`
+            const filename = response.headers['content-disposition'].split('filename\="')[1].split('\"')[0]
```
stanleypliu, line 46: Seems like this could benefit from more clarity with regards to what [1] and [0] are? I'm guessing `[1]` is the individual header value for the filename including the path and `[0]` is the filename itself?

lib/modules/download/generators/gdb.rb
```diff
@@ -19,17 +19,14 @@ def initialize zip_path, wdpa_ids
   def generate
     return false if @wdpa_ids.is_a?(Array) && @wdpa_ids.empty?
 
-    gdb_paths = []
-
     clean_up_after do
       QUERY_CONDITIONS.each do |name, props|
```
stanleypliu, line 23: Freeze `QUERY_CONDITIONS`?

lib/modules/download/generators/gdb.rb
```diff
@@ -70,20 +76,18 @@ def clean_up_after
...
   end
 
   def zip_path
     File.join(@path, "#{@filename}.zip")
   end
 
-  def gdb_component(name)
-    File.join(@path, "#{@filename}_#{name}.gdb")
+  def gdb_component
+    File.join(@path, "#{@filename}.gdb")
```
stanleypliu, line 87: Potentially you could condense  `zip_path` and `gdb_component` into one method to reduce the method count further and then simply pass a variable to determine the file type, what do you think?

lib/modules/ogr/postgres.rb
```diff
@@ -35,4 +40,41 @@ def self.ogr_command template, context
...
+    attrs = filename.split('_')
+    # Given the original filename should also contains 'polygons' or 'points' at the end,
+    # we remove this bit.
+    attrs.pop if %w(polygons points).include?(attrs[-1])
+    # Replace 'Public' with the geometry type formatted correctly, e.g. => ['WDPA', 'MmmYYY', 'poly']
+    attrs[-1] = FEATURE_TYPES[geom_type]
+    # Swap date and geometry type, e.g. => ['WDPA', 'poly', 'MmmYYY']
+    attrs[-1], attrs[-2] = attrs[-2], attrs[-1]
+    # Join and get the new filename, e.g. => 'WDPA_poly_MmmYYY'
+    attrs.join('_')
```
stanleypliu, line 78: Great explanation about the new naming conventions

lib/modules/download/generators/gdb.rb
Levia: Yeah that's an option. Although I think I would be more in favour of having 2 different methods as those are essentially two different files even if they share the same name. 
Also `zip_path` is shared across the parent-child structure of the generators and used in the base.rb file (the parent)

app/javascript/components/forms/DownloadCsv.vue
Levia: You are right, I've just added a comment that explains that. Thanks!

lib/modules/download/generators/gdb.rb
Levia: Well spotted. Done for both gdb and shapefile, thanks!

lib/modules/ogr/postgres.rb
stanleypliu: Just had a look though at Canada's GDB downloads as an example, and using Heather's comment in the Codebase ticket as context: 'The ISO3 code or WDPA ID can be added as a suffix where appropriate.' Shouldn't this be present at the end of the table name after 'Public'?

Levia: Ah well spotted. 'Public' should have gone too, so what I think it happens is that for downloads with an identifier at the end, it removes the identifier instead of 'Public'. I'll change that.
Thanks!

stanleypliu:
Everything looks and works as expected in terms of filenames, table names and so forth. Good job, merging

PR #592, opened 2021-01-27
Fix/download citations

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/189

Fixed download citations to show the month and year of the latest release, as well as updating the text of the citation to differ depending on the page that the download modal is currently open within. This differs for the WDPA page, OECM page and all other pages. 

⚠️ Also includes commits from `develop` which were not merged in previously - these relate to emergency fixes for the country and region maps on `master`. I have mentioned these in the release notes for 4.4.0


app/helpers/downloads_helper.rb
```diff
@@ -82,4 +85,20 @@ def download_text
       }
     }
   end
+
+  def switch_citation_text
+    # Should reflect the month/year of the last monthly release
+    current_month_year = Date.parse(Wdpa::S3.current_wdpa_identifier)
```
Levia, line 91: I would perhaps be a little bit more cautious with the identifier from S3.
It's quite unlikely that goes down, but if that happens or if the identifier returned is not in the expected format, `Date.parse` will raise an exception. 

To avoid that I would suggest either putting that value into a separate variable and check it later, or add a begin-rescue block rescuing for ArgumentError to catch exceptions related to Date.parse and manage this possibility.

Levia (changes requested):
Just a minor comment. Looks good otherwise!

app/helpers/downloads_helper.rb
stanleypliu: Good point...I might add the current month/year as a fallback (here's hoping that the releases are regular and won't be interrupted or delayed)

Levia: Uhm, I'm a bit undecided on that. Perhaps you could get any ProtectedArea object and get the month and year from the `created_at` date? `created_at` should reflect the release as those object are recreated each month.

stanleypliu: Actually before I change this, let me confirm with Heather whether this is advisable.

stanleypliu: Have been given permission to use current month and year.

Levia:
Works well. Merging!

PR #596, opened 2021-02-17
Fix/search accuracy

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/164

Search (navbar) improvements: 

- Ignores accents

Search reindexing improvements: 
- `rake search:reindex` does not break if the SLACK_WEBHOOK_URL env var isn't present 

Not fully tested for every combination of words and it also showed varying results when tested locally vs staging, but Heather has tested and confirmed that she's happy


lib/modules/search/index.rb
```diff
@@ -2,23 +2,30 @@ class Search::Index
...
     pa_relation = ProtectedArea.without_geometry.includes([
-      {:countries_for_index => :region_for_index},
-      :sub_locations,
-      :designation,
-      :iucn_category,
-      :governance
-    ])
-
-    region_index = Search::Index.new Search::REGION_INDEX, Region.without_geometry.all
+                                                            { countries_for_index: :region_for_index },
```
stacytalbot, line 17: Can you remove the giant indent in this file?

lib/modules/search/index.rb
```diff
@@ -80,21 +86,20 @@ def count
...
 
   def self.init_cms_index
     page_relation = Comfy::Cms::SearchablePage.includes([
-      :fragments_for_index,
-      {:translations_for_index => :fragments_for_index},
-      :categories
-    ])
-    Search::Index.new Search::CMS_INDEX, page_relation
+                                                          :fragments_for_index,
+                                                          { translations_for_index: :fragments_for_index },
```
stacytalbot, line 94: and here

PR #598, opened 2021-02-24
Fix/ticket 193

stanleypliu:
**TIcket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/193

I added safe navigation operators (the `&`) to the Country presenter's `build_stats` method when it fetches the PAME stats from the Statistic presenter as PAME statistics aren't available for every single national entity. This is a quick and dirty solution as there's other things that could be improved upon, but we want to get the pages for San Marino and the other three 'countries' working again quickly (Nauru, Vatican City and Macau).

### DIFF
--- a/app/controllers/country_controller.rb
+++ b/app/controllers/country_controller.rb
@@ -70,13 +70,12 @@ class CountryController < ApplicationController
   end
 
   def build_hash(tab)
-    oecm = tab == :wdpa_oecm
     hash = {}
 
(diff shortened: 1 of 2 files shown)


lucacug:
@stanleypliu As you mentioned in the comment, there will be the need for a refactor to the way we retrieve the pame statistics, but as quick solution I'm happy with this.

PR #603, opened 2021-03-30
Fix/show resource links and files

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/206 

- Restored lost styling of list items/links (use `button-download` mixin for `button--download`)
- Custom URL validation of resource link URLs, monkey patch CMS Pages controller to display custom error message upon submission of invalid URLs 
- Remove old code in CMS helper


stanleypliu:
Just want to make the custom validator less specific to Comfy so it can be reused elsewhere

stanleypliu:
@stacytalbot This can be reviewed now

PR #606, opened 2021-04-08
Fix/cta wiping

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/208

Disable CTA seeding during the WDPA release as it was preventing the CTAs already present in the database from being copied over to the new DB. 

Also contains a fix for the homepage stat cards' percentages. 

NB: There is a separate PR for the seeds as they are contained within the `db` submodule.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "db"]
   path = db
   url = https://github.com/unepwcmc/protectedplanet-db.git
-  branch = refresh
+  branch = fix/cta-wiping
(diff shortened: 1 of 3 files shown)


lucacug:
@stanleypliu  Happy with the code, did you manage to test it on staging and check this is actually working? If so, happy for this to be merged

stanleypliu:
Yes, tested this on staging and it works. I believe you may need to merge the PR in the submodule first, before this one

PR #607, opened 2021-04-08
Flags

stacytalbot:
⚠️ don't deploy to Prod until https://github.com/unepwcmc/ProtectedPlanet/pull/606 has been merged.

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/194


Joseph-Burke:
I can confirm that the broken flags are now working on this branch, but I have noticed that countries whose names contain spaces still have broken flags.

image

image

Besides this, #606 has not yet been merged, so unsure if this PR can be approved just yet.

> ⚠️ don't deploy to Prod until #606 has been merged.
> 
> https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/194

app/controllers/country_controller.rb
```diff
@@ -101,4 +101,11 @@ def load_essential_vars
     @country_presenter = CountryPresenter.new(@country)
     @tab_presenter = TabPresenter.new(@country)
   end
+
+  private 
+
+  def flag_path country_name
```
stanleypliu, line 107: I'm in favour of putting brackets around the argument just to make it more explicit what is being passed in.

Joseph-Burke: Will make this change now

app/controllers/country_controller.rb
```diff
@@ -105,7 +105,7 @@ def load_essential_vars
   private 
 
   def flag_path country_name
-    country_name = country_name.downcase.gsub(' ', '-')
-    ActionController::Base.helpers.image_url("flags/#{country_name}.svg")
+    file_string = country_name.downcase.gsub(' ', '-').gsub(',', '')
```
stanleypliu: Instead of chaining `.gsub`, what you could do is have a hash of characters to replace, like: 
```
characters_to_replace = {
 ' ' => '-',
',' => ''
}
```
and then loop through that hash, so something like:
```
name = country_name.downcase
characters_to_replace.each do |char, replacement| 
 name.gsub!(char, replacement)
end
```
so in future if there's a new character we need to replace (that's non-ASCII, you'll need I18n to replace that), we can just add it to the `characters_to_replace` hash

stanleypliu: Alternatively you can pass `gsub` a block and put conditional logic in there.

stanleypliu:
@Joseph-Burke I've tested a range of names (with accents, brackets and so on) and all work well. There is at least one particular country/territory though which 500s when it tries to access the flag for it: Heard Island and McDonald Islands. Do you want to have a look at why it's erroring?

Joseph-Burke:
It's showing up okay for me! The file should be found at:

/assets/flags/heard-island-and-mcdonald-islands-2a4b35ad48cfa6061dc49e9ef8c399e36f99572fe01bea273fd785a0c47ebb33.svg

image

PR #608, opened 2021-04-16
Map thumbnails

stacytalbot:
Fix up map thumbnails.

Issue was mainly due to us using an old classic style.

Not an obvious way to deal with the logo at this stage so leaving for now as it is more important to get the tiles up on prod.

N.B - New var needed for .env file


lucacug:
happy with this. Merging

PR #613, opened 2021-04-29
Add in new component that highlights countries that are using data f…

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/197

Add a message to country pages where the country is using restricted PAs to calculate the stats.

Each country has a different message.


Joseph-Burke:
Reviewing now

stanleypliu:
Can confirm that the messages appear for the required countries

PR #618, opened 2021-05-12
Fix/wrong data download

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/217

Fixed the areas search downloads not downloading more than 10000 results (the max page size set by Elasticsearch) if the specific filters/search returned more results than that number. 

Also added a task for clearing the Redis cache (which holds all of the download keys for each download) - useful if Sidekiq doesn't seem to work.

Test cases:
(no search term entered)
- Marine - true - should output around 19000 rows for CSV download
- Terrestrial - true - should be about 240000 rows

**Notes**

You may need to reindex your search before trying this out.


stanleypliu:
Jess has tested on staging and confirmed that she's happy for it to be merged.

PR #622, opened 2021-05-26
Add cookie banner using cookieconsent package

Joseph-Burke:
**This Pull Request addresses Protected Planet - Support & Maintenance [Ticket 163](https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/163), "Cookie Compliance Banner".**

I tried to use the `cookieconsent` gem at first, but came across some problems where Sprockets wasn't including it, so I went via the npm/webpacker route instead.

- `cookieconsent` package added via npm and imported in `vue.js`
- `_cookieconsent.html.erb` partial created to store the necessary script tag where the cookieconsent banner is initialised.
- That partial is then rendered at the top of the body in `application.html.erb`.

Note: There are lots of config options to customise the appearance and behaviour of the banner. I've left it quite simple for now and given the button a colour to match the colour scheme of the page.

image

### DIFF
--- a/app/views/layouts/partials/_cookieconsent.html.erb
+++ b/app/views/layouts/partials/_cookieconsent.html.erb
@@ -1,20 +1,22 @@
 <script>
-   cookieconsent.initialise({
-        palette: {
-            popup: {background: "#fff"},
-            button: {
-                background: "#71A32B",
-                text: '#fff'
-            },
-        },
-        revokable: true,
-        onStatusChange: function(status) {
-        console.log(this.hasConsented() ?
-        'enable cookies' : 'disable cookies');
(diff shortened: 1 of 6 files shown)


app/views/layouts/partials/_cookieconsent.html.erb
```diff
@@ -0,0 +1,20 @@
+<script>
+   cookieconsent.initialise({
+        palette: {
+            popup: {background: "#fff"},
```
stacytalbot: I'm just going to swap this to use the dark grey instead of white so it doesn't merge in with the rest of the page so much.

PR #623, opened 2021-05-27
Ticket 218/filepath too long

lucacug:
Codebase ticket  https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/218

When the full dataset gets downloaded from the search page https://pp.new-web.pp-staging.linode.protectedplanet.net/en/search-areas?geo_type=country without applying any filters or search term, an identifier (SHA256 hash) gets unnecessarily added as a suffix to the filename.
This is a hotfix to prevent this

### DIFF
--- a/lib/modules/download/utils.rb
+++ b/lib/modules/download/utils.rb
@@ -80,6 +80,7 @@ module Download
     end
 
     def self.search_token term, filters
+      return '' if (term.empty? && filters.empty?)
       Digest::SHA256.hexdigest(term.to_s + filters_dump(filters))
     end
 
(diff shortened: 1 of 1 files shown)


lib/modules/download/utils.rb
```diff
@@ -80,6 +80,7 @@ def self.filters_dump filters
     end
 
     def self.search_token term, filters
+      return 'all' if (term.empty? && filters.empty?)
```
stanleypliu, line 83: Do you know if we have access to the `geo_type` query param in this method? If so, instead of just passing `all` we could pass `all-#{geo_type.pluralise}` where it could be all countries, all regions etc. which would be more informative

stanleypliu:
Can confirm it works (after pushing the updated branch to staging) and also points are included in the downloads as well, and they are visible when opened in QGIS

PR #629, opened 2021-07-01
Fix/stats cards

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/224

Fix timeouts on country pages that can occur because the geometries of related sites (used for the example sites component towards the bottom of the page) are passed to the country pages. If the geometries are particularly complex, the page takes too long to load and times out. A new concern has been created specifically for this which allows scopes (or methods) to be passed to ProtectedArea ActiveRecord Collections without having to pass the geometry and other unused attributes which increases load times. 
Code responsible for the rendering of thumbnail links have also been fixed in the StatsSites Vue component.


app/models/concerns/relatable_protected_areas.rb
```diff
@@ -0,0 +1,22 @@
...
+    protected_areas.yield_self do |protected_areas|
+      block_given? ? yield(protected_areas) : protected_areas
+    end.limit(limit).order(:name)
+  end
+
+  def related_protected_areas_without_geometry(limit: nil)
+    protected_areas.without_geometry.yield_self do |protected_areas|
+      block_given? ? yield(protected_areas) : protected_areas
+    end.limit(limit).order(:name)
+  end
```
lucacug, line 21: Nice!!

lucacug (approved):
This is working well, I don't experienced the timeout on the DNM page and happy with the code BE wise.
@stacytalbot  Do you wanna have a quick look at the FE code? otherwise I reckon this can be merged in

PR #643, opened 2021-10-07
Fix/wdpa s3 file selection

defaye:
fix: select latest WDPA_WDOCEM more accurately
    
    select from S3 based on date in name
    rather than based on the last modified date


lib/modules/wdpa/s3.rb
```diff
@@ -35,7 +35,15 @@ def current_wdpa_identifier
...
-    available_wdpa_databases.sort_by(&:last_modified).last.object
+    latest = available_wdpa_databases.max_by do |object|
+      filename = object.key # e.g. "WDPA_WDOECM_Sep2020_Public.gdb.zip"
+
+      Date.parse(filename)
+    rescue ArgumentError
+      # Ignore files that cannot be parsed
+      Date.parse('1900-01-01')
+    end
+    latest.object
```
Joseph-Burke, line 46: Brilliant, love this code @defaye

Joseph-Burke (approved):
Makes sense, is legible, fixes our issues with this permanently 🤞🏻. Approved.

PR #670, opened 2022-06-06
feat: added a download link for global statistics

sergiomarrocoli:
Adds a download link for the global statistics csv

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/261

Testing:

go to the home page, click the download link

image

### DIFF
--- a/app/controllers/global_statistics_controller.rb
+++ b/app/controllers/global_statistics_controller.rb
@@ -1,14 +1,5 @@
 class GlobalStatisticsController < ApplicationController
   def show
-    send_file filename, type: 'text/csv', disposition: 'attachment'
-  end
-
-  private
-
-  def filename
-    directory = "#{Rails.root}/lib/data/seeds/"
-    global_statistics_csvs = Dir.glob("#{directory}/global_statistics*")
-    latest_global_statistics_file = global_statistics_csvs.sort.last
-    latest_global_statistics_file
+    send_file GlobalStatistic.latest_csv, type: 'text/csv', disposition: 'attachment'
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/global_statistics_controller.rb
```diff
@@ -0,0 +1,5 @@
+class GlobalStatisticsController < ApplicationController
+  def show
```
lucacug, line 2: can we rename this action to something like `download` ?

config/routes.rb
```diff
@@ -34,6 +34,8 @@
     get '/country/:iso/compare(/:iso_to_compare)', to: 'country#compare', as: 'compare_countries'
     get '/country/:iso/protected_areas', to: 'country#protected_areas', as: 'country_protected_areas'
 
+    get '/global_statistics', to: 'global_statistics#show'
```
lucacug, line 37: change this accordingly to the controller action name comment

lib/modules/wdpa/global_stats_importer.rb
```diff
@@ -10,7 +10,10 @@ def self.import
     CSV.foreach(latest_global_statistics_csv, headers: true) do |row|
       field = row['type']
       value = parse_value(row['value'])
-      attrs.merge!("#{field}": value)
+
+      # The global_statistics csv can now be downloaded so a methodology url has been added
+      # to the end of the spreadsheet. We need to not add this line to the attributes.
+      attrs.merge!("#{field}": value) if field.present?
```
lucacug, line 16: yeah it makes sense

lucacug (changes requested):
just a small name change, after that it is good to go

app/controllers/global_statistics_controller.rb
sergiomarrocoli: changed

config/routes.rb
sergiomarrocoli: changed

lucacug (approved):

PR #676, opened 2022-06-08
feat: Use hardcoded bounding boxes for countries that cross the dateline

sergiomarrocoli:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/166

Country show pages are not zooming to the correct extent, becaue the API we query doesn't handle countries that cross the dateline well, returning min -179.99... and max 179.99...

I spoke to Osgur about this and he suggested that the simplest thing would be if we could handle it on our end, rather than modifying the API. It only affects three countries, and as it's unlikely to change very often, i've added a method that returns hardcoded bounding boxes for these countries. if they are present in mapOptions they get used to initialize the map, if not the API is used.

Testing:
 go to 
`http://localhost:3000/country/NZL`
`http://localhost:3000/country/RUS`
`http://localhost:3000/country/FJI`
and check the bounding box looks ok.

go to other countries and check they display ok.

Draft because:
need to consider protected areas?
UNEP GIS API is down currently so can't test PAs and non dateline countries


sergiomarrocoli:
will be investigated and changes made on api if possible

PR #677, opened 2022-06-09
Feat/263 verifier field not showing

sergiomarrocoli:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/263

The verifier field is in our database, but not appearing in the sources csv returned with csv and shp file downloads. this adds it.

Testing:

The file is generated monthly and saved as a csv, so delete that if you have already downloaded this month. file is called e.g., WDPA_sources_May2021.csv

go to a protected area or country, download the csv and shapefile. sources table should include the verifier field


lucacug:
works as expected

PR #678, opened 2022-06-09
feat: added maxZoom: 10 to mapbox defaults, and overide in the protec…

sergiomarrocoli:
…ted area controller

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/230

Maps that show the cached webmap service tiles have an issue where the Protected Areas disappear beyond a zoom level (10), because they aren't cached. this is already a pretty close zoom, so for now i've limited the default zoom to 10. this can be overridden in the map_options object, and i've done so here for the protected area show page, because that loads a single polygon retrieved from a different source, so can be zoomed in to any level

Testing:
go to the different map pages (country, site, homepage, examples listed in the codebase ticket), and zoom in as far as possible.  PAs should not disappear.

### DIFF
--- a/app/controllers/protected_areas_controller.rb
+++ b/app/controllers/protected_areas_controller.rb
@@ -43,7 +43,8 @@ class ProtectedAreasController < ApplicationController
 
     @map_options = {
       map: {
-        boundsUrl: @protected_area.extent_url
+        boundsUrl: @protected_area.extent_url,
+        maxZoom: 0
       }
     }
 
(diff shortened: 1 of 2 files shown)


lucacug:
I have tested this on the country and global maps and it is working as intended. While testing the single PAs show page, this is how the map looked like and the error is the following:
```
{
 "error": {
  "code": 400,
  "extendedCode": -2147220985,
  "message": "Unable to complete operation.",
  "details": [
   
  ]
 }
}
```
This is not related to this specific PR as I can see the same behavior on prod (something to double check asap I would say), but just mentioning here as I wasn't able to test the zoom level 0

lucacug:
same on production, jus
Screenshot from 2022-06-15 23-54-56
t for reference

sergiomarrocoli:
fix for the individual PAs is here: https://github.com/unepwcmc/ProtectedPlanet/pull/688

lucacug:
this work as expected, merging

PR #679, opened 2022-06-09
Use correct bucket for file downloads

sergiomarrocoli:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/272

Second attempt at fixing this bug.

File downloads always point at the latest version of the database in the s3 buckets. Last attempt to fix it broke because the monthly import looks in the same place as the file downloads for the file, so it was only finding the current release's file using the constants in the constants.rb, not the newer one.

This points file downloads to the current release files and the monthly import to the current release. 

Testing:
I found it easiest to update the secrets file so all development credentials point to the production buckets, then:
download a csv and shapefile for a country (jordan and Tanzania had changes recently so is a good choice)- should be the latest bucket (currently Jun 2022).
update the constants to an earlier date (for Jordan and Tazania the changes appeared in Mar 2022 release, so before that) 
restart sidekiq and the rails server
redownload the shp file and csv for the same country
compare the files, later files should be bigger

I haven't tested the full import yet

### DIFF
--- a/lib/modules/wdpa/release.rb
+++ b/lib/modules/wdpa/release.rb
@@ -14,7 +14,7 @@ class Wdpa::Release
   end
 
   def download
-    Wdpa::S3.download_current_wdpa_to zip_path
+    Wdpa::S3.download_latest_wdpa_to zip_path
     system("unzip -j '#{zip_path}' '\*.gdb/\*' -d '#{gdb_path}'")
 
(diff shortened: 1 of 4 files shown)


lucacug (approved):
code looks good, mergin this now and testing the download as the PP team has uploaded the new gdb on S3. We will test the release import shortly as well

PR #683, opened 2022-06-10
feat: Use latest PAME data csv for date of generated download csv

sergiomarrocoli:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/268

Uses the last PAME data file to set downloaded file name instead of a constant

### DIFF
--- a/app/models/pame_evaluation.rb
+++ b/app/models/pame_evaluation.rb
@@ -340,7 +340,7 @@ class PameEvaluation < ApplicationRecord
   def self.last_csv_update_date
     pame_data_csvs = Dir.glob("#{Rails.root}/lib/data/seeds/pame_data*")
     latest_pame_csv = pame_data_csvs.sort.last.split('_').last
-    latest_pame_csv.split.first.to_date
+    latest_pame_csv.split('.').first.to_date
   end
 end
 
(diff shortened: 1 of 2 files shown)


app/models/pame_evaluation.rb
```diff
@@ -336,6 +336,12 @@ def self.to_csv(json = nil)
 
     generate_csv(where_statement, restricted_where_statement)
   end
+
+  def self.last_csv_update_date
+    pame_data_csvs = Dir.glob("#{Rails.root}/lib/data/seeds/pame_data*")
+    latest_pame_csv = pame_data_csvs.sort.last.split('_').last
+    latest_pame_csv.split.first.to_date
```
lucacug, line 343: this is working well, but the `split.first` is having no effect on the string. Either we remove that bit (apparently it is smart enough to convert something like `"2022-03-01.csv"`) into a date anyway or we specify the splitter symbol (in this case `.`)

lucacug (changes requested):
nice refactor here, the update constant bit was part of the wdpa release (it might have gone missing along the way at some point) but this automation will make our life easier. Good stuff.
Just a minor comment to address and it is good to go

app/models/pame_evaluation.rb
sergiomarrocoli: added '.' pattern as split argument

PR #684, opened 2022-06-10
fix: give correct link for regions in filter results

sergiomarrocoli:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/273

Regions returned by filter results did not link correctly to the region page

Testing:
search for a region (north america, europe, africa etc)
link should take you to that page

NOTE: bounding box isn't selected properly. think this should be fixed in another ticket where will try to pass up bounding boxed using our own database, because the bounding boxes from the api we use has issues (e.g. dateline bounding boxes fail)


lucacug:
Working as expected, I will check the bounding boxes in the other PR when it is ready

PR #688, opened 2022-06-16
fix/Changed urls so they work after recent API changes

sergiomarrocoli:
https://unep-wcmc.codebasehq.com/projects/gef-spatial/tickets/1

Some api requests are broken after recent API changes. they work again if we submit a number instead of a string


lucacug:
this is working now, thanks for the fix, I would suggest making this live asap.
A bit worrying though, do you have the link where they mention this API change? We were lucky was a quick win but they really should versioning the API or properly advertise any breaking change

PR #689, opened 2022-06-16
feat: moved download link to same line as latest update

sergiomarrocoli:
Moved download link adjacent to latest file info as requested by Charlotte

http://localhost:3000/en

image


lucacug:
looks good

PR #692, opened 2022-06-17
fix: country zoom url

sergiomarrocoli:
fixes broken country url

got to home page, enter country into search on map
should zoom to country
same for region, e.g. Latin America & Caribbean

### DIFF
--- a/app/helpers/map_helper.rb
+++ b/app/helpers/map_helper.rb
@@ -104,7 +104,7 @@ module MapHelper
 
   def country_extent_url (iso3)
     {
-      url: "https://data-gis.unep-wcmc.org/server/rest/services/GADM_EEZ_Layer/FeatureServer/0/query?where=iso_ter+%3D+#{iso3}&returnGeometry=false&returnExtentOnly=true&outSR=4326&f=pjson",
+      url: "https://data-gis.unep-wcmc.org/server/rest/services/GADM_EEZ_Layer/FeatureServer/0/query?where=iso_ter+%3D+%27#{iso3}%27&returnGeometry=false&returnExtentOnly=true&outSR=4326&f=pjson",
       padding: 5
     }
   end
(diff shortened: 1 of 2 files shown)


app/helpers/map_helper.rb
```diff
@@ -104,7 +104,7 @@ def all_services_for_point_query
 
   def country_extent_url (iso3)
     {
-      url: "https://data-gis.unep-wcmc.org/server/rest/services/GADM_EEZ_Layer/FeatureServer/0/query?where=iso_ter+%3D+#{iso3}&returnGeometry=false&returnExtentOnly=true&outSR=4326&f=pjson",
+      url: "https://data-gis.unep-wcmc.org/server/rest/services/GADM_EEZ_Layer/FeatureServer/0/query?where=iso_ter+%3D+%27#{iso3}%27&returnGeometry=false&returnExtentOnly=true&outSR=4326&f=pjson",
```
lucacug, line 107: ok so if I understand correctly, based on the type of service (Map or Feature) the query string requires strings or numbers as params?

sergiomarrocoli: I'm not sure.

the Protected area extent and boundaries use Map with a single wdpaid, which can't be a string.

Country extent does use a feature endpoint, but it's not querying on wdpaid, but iso_ter, so it was a mistake to remove the quotations here. Region uses a map endpoint, but again not wdpaid. these end points don't take a wdpa_id, at all

lucacug: Thanks Sergio! I was more thinking for future-proofing would be good to understand when we need to send a string or an integer in the query string due to this API change on the ESRI side.
In this case, the fix works so going to merge, but let's have a chat about this, maybe including Osgur in the loop

lucacug (approved):

PR #693, opened 2022-06-27
Fix/261 add back file with additional column

sergiomarrocoli:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/261

The global statistics file is now available to download is not the same as the one originally supplied by the PP team, because a new 'description' column was removed after the import failed. The failure was not due to the new column, but a trailing methodology link. 

The importer was updated to allow it to import the original file in an earlier PR, but the original file was not restored, so this does that.

Testing:
go to the homepage, click the link to download the global statistics
it should have a 'description' column, and a methodology link in the centre column at the bottom of the sheet


sergiomarrocoli:
Not needed, release in two days

PR #698, opened 2022-08-01
fix: update designations presenter and country and region classes to …

sergiomarrocoli:
…return 'Not Applicable' designations along with 'National' designations

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/241

before: PAs designated as 'Not Applicable' were not included in the Designations tables in the country and region show pages

after: they are included in the 'National' designations.

changes: 
- updated the ```protected_areas_per_designation``` method in country and region to take an array of designations instead of one
- updated the SQL method in ```protected_areas_per_designation``` to use IN instead of =, and to group by designation name (because there are separate records for e.g. Marine Protected Area for each designation type, i.e. 'Not Applicable', 'National' etc 
- updated the presenter to pass a list of jurisdictions to ```protected_areas_per_designation```, adding 'Not Applicable' when jurisdiction == 'National'
- used the response from ```protected_areas_per_designation``` to generate count and % statistics, rather than having to hack the existing methods

Testing:
- go to pages where countries have some 'Not Applicable' PAs, e.g. Samoa http://localhost:3000/country/WSM
- On this branch the total at the top of the page should match the sum of the different jurisdictions
image
image

- on develop the above should not match
- on this branch Marine protected area count should be higher than on develop
- should work the same for regions

### DIFF
--- a/app/models/country.rb
+++ b/app/models/country.rb
@@ -98,15 +98,16 @@ class Country < ApplicationRecord
     convert_into_hash(sources.uniq)
   end
 
-  def protected_areas_per_designation(jurisdiction=nil, exclude_oecms: false)
+  def protected_areas_per_designation(jurisdictions=[], exclude_oecms: false)
     ActiveRecord::Base.connection.execute("""
-      SELECT designations.id AS designation_id, designations.name AS designation_name, pas_per_designations.count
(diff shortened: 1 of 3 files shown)


lucacug:
Code looks good and works as expected (I only tested the Samoa case, but we can ask the NC team to double check). Merging!

PR #699, opened 2022-08-01
fix: replace the call to OECM feature end point with requests to the …

sergiomarrocoli:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/286

Clicking on OECMs on the map on the home page didn't work

this PR changes the FeatureSerice endpoint we were querying for two MapServer endpoints (polygon and point)

Testing:
On home page, click on different PA types, each should cause a popup, and the link should take you to the show page for that PA

### DIFF
--- a/app/helpers/map_helper.rb
+++ b/app/helpers/map_helper.rb
@@ -79,9 +79,10 @@ OVERLAYS = [
 ].freeze
 
 ALL_SERVICES_FOR_POINT_QUERY = [
-  { url: OECM_FEATURE_SERVER_LAYER_URL, isPoint: false },
+  { url: WDPA_POLY_LAYER_URL, isPoint: false },
   { url: WDPA_POINT_LAYER_URL, isPoint: true },
-  { url: WDPA_POLY_LAYER_URL, isPoint: false }
+  { url: OECM_POINT_LAYER_URL, isPoint: true },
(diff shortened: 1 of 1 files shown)


Joseph-Burke:
Working on a fix for this but just recording here that the README's set-up instructions are not working in this branch due due to errors during dependency installation:

`docker-compose build` -->
```
#12 78.28 
#12 78.28 An error occurred while installing mimemagic (0.3.10), and Bundler cannot
#12 78.28 continue.
#12 78.28 Make sure that `gem install mimemagic -v '0.3.10' --source
#12 78.28 'https://rubygems.org/'` succeeds before bundling.
#12 78.28 
#12 78.28 In Gemfile:
#12 78.28   comfortable_mexican_sofa was resolved to 2.0.19, which depends on
#12 78.28     comfy_bootstrap_form was resolved to 4.0.7, which depends on
#12 78.28       rails was resolved to 5.2.0, which depends on
#12 78.28         activestorage was resolved to 5.2.0, which depends on
#12 78.28           marcel was resolved to 0.3.3, which depends on
#12 78.28             mimemagic
```

Joseph-Burke (approved):
Haven't been able to test the effect of these changes due to setup issues, but this seems fine.

PR #703, opened 2022-08-03
feat: add greenlist candidate filter and update filter presets and bu…

sergiomarrocoli:
…tton text
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/235

Adds an extra greenlist filter, and auto-selects them when clicking into the search via the link on the greenlist page

original description in ticket includes adding fields to csv export, but this is not now needed.


docs/search.md
```diff
@@ -99,3 +99,34 @@ Run this in the console:
...
+} AND {
+  nested_field_1: [nested_filter_1a OR nested_filter_1b OR nested_filter_1c]
+} AND {
+  nested_field_2: [nested_filter_2a OR nested_filter_2b OR nested_filter_2c]  
+}
+```
+
+***However*** Because of the way the filters are set up in the UI, boolean filters appear to be processed as OR statements. They are not, we delete some of the filters on the back end to make it appear this way. If a user selects WDPA and OECM, if we submitted both of those parameters we would get zero results, because a Protected Area can't be both of them. so in the back end we delete the filters if both are submitted. same for Type. This works ok, because both categories are binary, and the WDPA/OECM and Marine/Terrestrial categories aren't mutually exclusive. If you have a filter that isn't binary, e.g. Green Listed, Green List Candidate, and Neither, we can't use this method.
+
+However, with Elasticsearch we can convert these binary values into nested values to make groups of non-exclusive binary options (see ProtectedArea#as_indexed_json and #special_status).
```
Joseph-Burke, line 132: 👌🏻

Joseph-Burke (approved):
Again, I haven't been able to verify the effect of these changes due to setup issues, but I've taken a look at the changes and the code looks good.

PR #705, opened 2022-08-03
fix: reduce complexity of geojson before sending to mapbox in uri, to …

sergiomarrocoli:
…prevent 403 error from mapbox

https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/131
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/203
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/271

mapbox was rejecting requests becuase the geojson in the uri made it too long (somewhere over 5000 characters), so thumbnails weren't being generated for PAs with complex geometries

this simplifies the geometry before sending if the PAs geometry is too complex and, if that doesn't simplify enough, uses a very simple outline simplification. 

testing:
put a byebug on line 11 of the AssetGenerator class (which will be hit if any of the tile requests fails)
go to ```http://localhost:3000/en/search-areas?search_term=wales&geo_type=site```
all thumbnails should load
go to http://localhost:3000/en/search-areas?filters%5Bdb_type%5D%5B%5D=wdpa and keep scrolling and let the thumbnails load, should not hit the byebug

### DIFF
--- a/app/models/concerns/geometry_concern.rb
+++ b/app/models/concerns/geometry_concern.rb
@@ -22,14 +22,67 @@ module GeometryConcern
     ]
   end
 
+  def geojson_for_mapbox_uri geo_properties=nil
+    # Returns a stringified geojson for making a request to the Mapbox API, to generate a preview thumbnail.
+    # If the string is too long it will be rejected, so we then need to simplify it.
+    # First we can buffer it, which reduces complexity but retains the original shape well.
+    # Finally, if the string is still too long, we fall back on a convex transformation, which
+    # provides a very crude outline, but is a short string.
+    if geojson_suitable_for_mapbox_url?((geojson = geojson_query(geo_properties)))
+      geojson
+    elsif geojson_suitable_for_mapbox_url?((buffered_geojson = buffered_geojson_query(geo_properties)))
+      buffered_geojson
+    else
+      convex_geojson_query(geo_properties)
+    end
+  end
+
+  def geojson_suitable_for_mapbox_url?(geojson)
+    geojson.present? && geojson.length <= 5000
+  end
+
   def geojson geo_properties=nil
-    geojson = ActiveRecord::Base.connection.select_value("""
-      SELECT ST_AsGeoJSON(ST_SimplifyPreserveTopology(ST_MakeValid(#{main_geom_column}), 0.003), 3)
+    geojson_query(geo_properties)
+  end
+
+  private
+
+  def buffered_geojson_query geo_properties
+    # We send a geojson to mapbox to create thumbnails. If the length of these exceeds a uri
+    # length of ~5000 characters, mapbox returns a 403. This method simplifies the geojson by
+    # first simplifying the geometry (to make the buffer faster), adding a buffer, and then
+    # simplifying again.
+    simplified_geojson = ActiveRecord::Base.connection.select_value("""
+    SELECT ST_AsGeoJSON(ST_SimplifyPreserveTopology(ST_Buffer(ST_SimplifyPreserveTopology(ST_MakeValid(#{main_geom_column}), 0.005), 0.005), 0.005), 3)
       FROM #{self.class.table_name}
       WHERE id = #{id}
     """.squish)
 
-    return nil unless geojson.present?
+    simplified_geojson.present? ? to_uri(simplified_geojson, geo_properties) : nil
+  end
+
+  def convex_geojson_query geo_properties
+    # Very simple polygon, if other polygon thumbnail methods fail.
+    simplified_geojson = ActiveRecord::Base.connection.select_value("""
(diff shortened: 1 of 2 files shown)


app/models/concerns/geometry_concern.rb
```diff
@@ -22,14 +22,55 @@ def bounds
     ]
   end
 
+  def geojson_for_mapbox_uri geo_properties=nil
+    geojson_uri = geojson_query(geo_properties)
+    geojson_uri = buffered_geojson_query(geo_properties) if geojson_uri.nil? || geojson_uri.length > 5000
+    geojson_uri = convex_geojson_query(geo_properties) if geojson_uri.nil? || geojson_uri.length > 5000
+    geojson_uri
```
Joseph-Burke, line 29: Is this correct? It seems like the assignments on 27 and 28 have identical conditions, so isn't the assignment on 28 always going to overwrite 27?

In any case, maybe this could be DRYed by storing `geojson_uri.nil? || geojson_uri.length > 5000` in a descriptive variable if we intend to use it more than once. Or if we're reusing the same condition, it might make more sense just to open a block instead of single-line syntax

app/models/concerns/geometry_concern.rb
```diff
@@ -22,14 +22,55 @@ def bounds
...
+    geojson_query(geo_properties)
+  end
+
+  private
+
+  def buffered_geojson_query geo_properties
+    # We send a geojson to mapbox to create thumbnails. If the length of these exceeds a uri
+    # length of ~5000 characters, mapbox returns a 403. This method simplifies the geojson by
+    # first simplifying the geometry (to make the buffer faster), adding a buffer, and then
+    # simplifying again.
```
Joseph-Burke, line 42: Great explanation 👍🏻

Joseph-Burke (changes requested):
See my comment on `geometry_concern.rb` below. Perhaps I'm misunderstanding something, but it seems like it could be a mistake

I don't know Mapbox super well, but your explanation of what's being done here is clear.

app/models/concerns/geometry_concern.rb
sergiomarrocoli: the value of geojson changes with each line, as it is reassigned, so although we check it's value in the same way ```geojson_uri.nil? || geojson_uri.length > 5000```, the value can be different. we want to check each tme, rather than calculating them all and then picking one, because they are relatively slow calculations. so line 27 is only called if the first attempt results in a json that is too long, and line 28 only gets called if the result from 27 is too long. 

i will move the presence and length check to a method. but not sure any rewrites make it much clearer, e.g.:
```
def geojson_suitable_for_mapbox_url?(geojson)
  geojson_uri.present? && buffered_geojson_uri.length <= 5000
end

# original with a method call to check presence and length
def geojson_for_mapbox_uri geo_properties=nil
  geojson_uri = geojson_query(geo_properties)
  geojson_uri = buffered_geojson_query(geo_properties) unless geojson_suitable_for_mapbox_url?(geojson)
  geojson_uri = convex_geojson_query(geo_properties) unless geojson_suitable_for_mapbox_url?(geojson)
  geojson_uri
end

# renaming geojson variable
def geojson_for_mapbox_uri geo_properties=nil
  geojson_uri = geojson_query(geo_properties)
  return geojson if geojson_suitable_for_mapbox_url?(geojson)

  buffered_geojson_uri = buffered_geojson_query(geo_properties)
  return buffered_geojson if geojson_suitable_for_mapbox_url?(buffered_geojson)

  convex_geojson_query(geo_properties)
end

# elsif statement
def geojson_for_mapbox_uri geo_properties=nil
  if geojson_suitable_for_mapbox_url?((geojson_uri = geojson_query(geo_properties)))
    geojson
  elsif geojson_suitable_for_mapbox_url?((buffered_geojson = buffered_geojson_query(geo_properties))))
    buffered_geojson
  else
    convex_geojson_query(geo_properties)
  end
end
```

Joseph-Burke: I think on balance, my favourite option is using the `geojson_suitable_for_mapbox_url?` method in combination with `if`/`elsif`/`else` and implicit returns. _But_, I only offer that as a bit of feedback from someone not so familiar with this codebase. I think that little helper method is nice because it wraps that conditional in a more semantic form. But please feel free to make the final call here, as I think you're more familiar with the code than I am.

In any case, it would be great to add a brief comment to help any future developer that comes across that repeat assignment stuff, to explain as you have here that the value of geojson changes.

And lastly, watch out if you use the code snippet above, I think `geojson_suitable_for_mapbox_url?` has an unused parameter as it's written there :)

Feel free to merge this PR once you make a call and add a little comment, thanks for the clarification 🙏🏻

PR #726, opened 2022-11-10
fix: pass the correct point query urls to each map component

WillKocur:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/286


roshan-mathewtech (commented):
Looks fine. Approved

PR #754, opened 2022-12-21
WDPA Jan 2023 release

pazolka:


sergiomarrocoli (approved):
looks good and importers run fine locally. approving, but will merge in Jan (and need to add to changelog, when we know which version it will be)

PR #764, opened 2023-02-13
fix: fix multipoints in .gdb downloads

pazolka:


lucacug (commented):
I suggest we test this on staging first, could we deploy this branch, clear the staging downloads S3 buckets pp-downloads-staging and run few downloads, please?

PR #773, opened 2023-03-09
feat: prohibit list of PAs by wdpa_id

sergiomarrocoli:
https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/326

Emergency fix which raises an error in the PA controller if a user tries to load a PA with a wdpa_id that is in a list provided by the PP team


pazolka (approved):

sergiomarrocoli:
short term fix, no longer needed

PR #774, opened 2023-03-09
feat: add api as a service in docker-compose.yml

sergiomarrocoli:
I added this to the .env file in protectedplanet-api to get it to work
```
APPSIGNAL_LOG_PATH=/var/log/appsignal/protectedplanet-api.log
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -97,7 +97,10 @@ services:
     command: /bin/bash -l -c "bundle exec rackup"
     volumes:
       - ${API_PATH}:/ProtectedPlanetApi
+      - ${SSH_AUTH_SOCK}:${SSH_AUTH_SOCK}
     network_mode: host
+    environment:
+      - SSH_AUTH_SOCK=${SSH_AUTH_SOCK}
     env_file:
       - ${API_PATH}/.env
     depends_on:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -90,6 +90,25 @@ services:
       - protectedplanet_bundler:/usr/local/bundle
     network_mode: host
 
+  api:
+    container_name: protectedplanet-api
+    build: ${API_ABSOLUTE_PATH}
+    command: /bin/bash -l -c "bundle exec rackup"
+    volumes:
+      - ${API_ABSOLUTE_PATH}:/ProtectedPlanetApi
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -90,6 +90,25 @@ services:
...
+      - ${SSH_AUTH_SOCK}:${SSH_AUTH_SOCK}
+    network_mode: host
+    environment:
+      - SSH_AUTH_SOCK=${SSH_AUTH_SOCK}
+    env_file:
+      - ${API_ABSOLUTE_PATH}/.env
+    depends_on:
+      - db
+    stdin_open: true
+    tty: true
```
sergiomarrocoli, line 108: not sure what these do

docker-compose.yml
```diff
@@ -90,6 +90,25 @@ services:
...
+    build: ${API_ABSOLUTE_PATH}
+    command: /bin/bash -l -c "bundle exec rackup"
+    volumes:
+      - ${API_ABSOLUTE_PATH}:/ProtectedPlanetApi
+      - ${SSH_AUTH_SOCK}:${SSH_AUTH_SOCK}
+    network_mode: host
+    environment:
+      - SSH_AUTH_SOCK=${SSH_AUTH_SOCK}
+    env_file:
+      - ${API_ABSOLUTE_PATH}/.env
```
sergiomarrocoli, line 104: what does this do?

docker-compose.yml
pazolka: needed for debugging, interactive console

docker-compose.yml
pazolka: mapping the ssh agent that was passed in the environmental variable right before `docker compose up` - not needed for the api service

pazolka: both lines 99 and 102 were needed to make the ssh agent work inside the container

docker-compose.yml
```diff
@@ -90,6 +90,25 @@ services:
       - protectedplanet_bundler:/usr/local/bundle
     network_mode: host
 
+  api:
+    container_name: protectedplanet-api
+    build: ${API_ABSOLUTE_PATH}
```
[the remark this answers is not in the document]
sergiomarrocoli: i think it does if it's outside the context. will check again

pazolka (approved):

PR #801, opened 2024-03-26
Chore/wdpa release 2024 04

WCMC-vblanque:


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-### 4.10.5
+### 4.10.6
```
IMDaniyal, line 1: You need to add the new changes for the April month. And keep the march changelog as well

WCMC-vblanque: Ok, now it is done I think.

PR #826, opened 2025-02-03
Chore/remove devise

yuelongh:
- remove devise gem and all related pages as it is not used anywhere in the app
- this PR include one migration to delete users table
- no current_user, user_signed_in, authenticate_user, has_secure_password, database_authenticatable, :registerable, :recoverable, :rememberable, :validatable found in the codes


IMDaniyal (approved):
Looks good

PR #828, opened 2025-03-06
Feat/add search whs filter

yuelongh:
- Add WHS filter to PA search page
- add debounce to search PA bar
- add ability to cms_serializer.rb to enable sorting by either default order or published_date order
- update layout for resources and WDPA release news template
- add appsignal


IMDaniyal (commented):
Apart from that one addition in ignore_errors, looks good 👍

IMDaniyal (approved):

PR #829, opened 2025-03-06
fix: remove .pdf accessibility

yuelongh:
- remove pdf acces to a PA to prevent website crashes as the function is not used anymore


IMDaniyal (approved):
All good 👍

PR #836, opened 2025-05-14
fix: apply filter to PAME list when a country has no PAs

yuelongh:
Currently, if a country i,e VAT has no PAs then it is returning all PAME list, so we need to catch this scenario.

[ticket](https://unep-wcmc.codebasehq.com/projects/protected-planet-support-and-maintenance/tickets/351)

### DIFF
--- a/app/models/pame_evaluation.rb
+++ b/app/models/pame_evaluation.rb
@@ -79,8 +79,6 @@ class PameEvaluation < ApplicationRecord
   end
 
   def self.parse_filters(filters)
-    site_ids = []
-    country_ids = []
     where_params = { sites: '', methodology: '', year: '', iso3: '', type: '' }
     filters.each do |filter|
       options = filter['options']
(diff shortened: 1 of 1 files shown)


app/models/pame_evaluation.rb
```diff
@@ -80,13 +81,15 @@ def self.generate_query(page, filter_params)
...
     filters.each do |filter|
-      options = filter["options"]
+      options = filter['options']
       case filter['name']
       when 'iso3'
         countries = options
-        site_ids << countries.map{ |iso3| Country.find_by(iso_3: iso3).protected_areas.pluck(:id) }
+        site_ids = countries.flat_map do |iso3|
+          Country.find_by(iso_3: iso3)&.protected_areas&.pluck(:id) || []
+        end
```
IMDaniyal, line 92: This site_ids = countries... will select only save the last iso3 counrteis incase multiple iso3s are selected for country filter. This will work here:

`site_ids += countries.flat_map....`

IMDaniyal (approved):

PR #858, opened 2026-04-14
Feat update marine page & Feat/hook pame gl from DMP

yuelongh:
- Add views for GL and PAME from DMP
- Update importers to use the views
- update frontend for  PAME, GL related 
- Update download site_id collector to depend on filter types so it doesn't always append all site_ids in the DB for some filtering.
- Update marine page

Chores:
- Remove API v3 as it shouldn't be there
- Remove networks as it is no longer used


app/models/pame_evaluation.rb
```diff
@@ -67,109 +73,93 @@ def self.structure_data(page, items)
...
+        where_params[:method] = quoted.empty? ? nil : "pame_methods.name IN (#{quoted.join(',')})"
       when 'year'
-        where_params[:year] = options.empty? ? nil : "pame_evaluations.year IN (#{options.join(',')})"
-      else
-        where_params[:type] = nil if options.empty? || options.length == 2
+        where_params[:year] = options.empty? ? nil : "pame_evaluations.asmt_year IN (#{options.join(',')})"
+      when 'type'
+        if options.length == 1
+          where_params[:type] = if options[0] == 'Marine'
+                                  'COALESCE(protected_areas.marine, protected_area_parcels.marine)'
```
IMDaniyal, line 108: 'COALESCE(protected_areas.marine, protected_area_parcels.marine) = true'

PostgreSQL will coerce the COALESCE expression as a truthy non-nil value regardless of the actual marine flag, so this filter has no effect.

app/presenters/protected_area_presenter.rb
```diff
@@ -147,33 +154,42 @@ def parcel_oecm_attributes(parcel)
...
-    }
+    protected_area.parcels_including_protected_area_self.sort_by(&:site_pid).map do |parcel|
+      {
+        site_pid: parcel.site_pid,
+        affiliation: 'parcc_info',
+        image_url: ActionController::Base.helpers.image_url('logos/parcc.png'),
+        link_title: "View the climate change vulnerability assessments for #{protected_area.name}",
+        link_url: url_for_related_source('parcc_info', protected_area)
+      }
+    end
```
IMDaniyal: has_parcc_info is a PA-level flag, not per-parcel, so mapping over all parcels produces N identical PARCC entries — one per parcel. The frontend groups affiliations by site_pid, so a PA with 3 parcels would show 3 PARCC logos in the PDF view.
Suggested fix — return a single entry keyed to the PA's own site_pid:

```
[{
  site_pid: protected_area.site_pid,
  affiliation: 'parcc_info',
  image_url: ActionController::Base.helpers.image_url('logos/parcc.png'),
  link_title: "View the climate change vulnerability assessments for #{protected_area.name}",
  link_url: url_for_related_source('parcc_info', protected_area)
}]
```

app/controllers/green_list_controller.rb
```diff
@@ -54,32 +54,27 @@ def map_overlays
   end
 
   def point_query_services
+    site_ids = pas_with_green_list_on_self_only.pluck(:site_id)
     wdpa_services_for_point_query.map do |service|
       service.merge({
-        queryString: site_ids_where_query(green_list_areas.map(&:site_id))
+        queryString: site_ids_where_query(site_ids)
       })
     end
```
IMDaniyal: pas_with_green_list_on_self_only only returns PAs where the GL status sits on the PA record. Any PA where the green list is held on a parcel only (parcel-level GL) won't appear in the map point query.
ProtectedAreaParcel.greenlisted_parcels was added in this PR — its site_id values can be unioned in directly:

```
def point_query_services
  pa_site_ids     = pas_with_green_list_on_self_only.pluck(:site_id)
  parcel_site_ids = ProtectedAreaParcel.greenlisted_parcels.pluck(:site_id)
  all_site_ids    = (pa_site_ids + parcel_site_ids).uniq

  wdpa_services_for_point_query.map do |service|
    service.merge({ queryString: site_ids_where_query(all_site_ids) })
  end
end
```

Also worth noting: the same gap affects terrestrial_green_list_area_ids (line 86) and marine_green_list_area_ids (line 90), which feed the terrestrial/marine map overlays via map_overlays. Those would need the same union treatment.

IMDaniyal (changes requested):

app/models/pame_evaluation.rb
yuelongh: If a type filter is present and Marine is chosen, then it filters out those entries that return FALSE or NULL. 
If a type filter is present but Marine isn't chosen, then it filters out those that return true and it will never be null so I think we are safe here?

IMDaniyal: Ah yes, that would work.

app/controllers/green_list_controller.rb
yuelongh: changed all to support any parcels specifically, not just when the first parcel that is gl

IMDaniyal (approved):

